PR #4, opened 2023-02-20
Hydrate project with copier

delucchi-cmu:


src/lsdb/example_module.py
```diff
@@ -0,0 +1,23 @@
...
+
+def meaning() -> int:
+    """The meaning of life, the universe, and everything.
+
+    Returns
+    -------
+    int
+        The meaning of life.
+    """
+    return 42
```
smcguire-cmu, line 23: Do we need this?

tests/lsdb_tests/test_example_module.py
```diff
@@ -0,0 +1,13 @@
...
+def test_greetings():
+    """Verify the output of the `greetings` function"""
+    output = example_module.greetings()
+    assert output == 'Hello from LINCC-Frameworks!'
+
+
+def test_meaning():
+    """Verify the output of the `meaning` function"""
+    output = example_module.meaning()
+    assert output == 42
```
smcguire-cmu, line 13: Same here, do we want this included?

smcguire-cmu (commented):
LSDB getting code, nice!

src/lsdb/example_module.py
delucchi-cmu: This is useful to have *something* tested in the workflows, even if it's meaningless.

I think I'd prefer something obviously meaningless to something that looks like it might be right but isn't (*cough* ChatGPT *cough*).

smcguire-cmu: Haha, fair enough

smcguire-cmu (approved):

PR #10, opened 2023-03-23
Updating to v1.3.0

delucchi-cmu:
All automated changes.


pyproject.toml
```diff
@@ -21,16 +22,15 @@ dependencies = [
 # On a mac, install optional dependencies with `pip install '.[dev]'` (include the single quotes)
 [project.optional-dependencies]
 dev = [
+    "mypy", # Used for static type checking of files
+  "pre-commit", # Used to run checks before finalizing a git commit
```
OliviaLynn, line 26:
```suggestion
    "pre-commit", # Used to run checks before finalizing a git commit
```

OliviaLynn: Another indentation thing--I'll go in to the template to check these over afterwards

tests/lsdb/conftest.py
OliviaLynn, line 1: Is this to provide some kind of coverage for Sphinx's conf.py/is it supposed to be blank for now?

delucchi-cmu: This gets generated by the template, but is a pytest thing and basically tells pytest "here be unit tests"

docs/Makefile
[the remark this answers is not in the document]
delucchi-cmu: We probably want notebooks in here eventually, so there's no harm here. It might make sense to drop from the template, though.

PR #17, opened 2023-08-09
Merge prototype to main

smcguire-cmu:
Code here is unit tested and already reviewed in #12 and #18, but was in the prototype branch because we weren't sure on the API interface yet.


src/lsdb/io/parquet_io.py
```diff
@@ -0,0 +1,29 @@
+import pandas as pd
+from hipscat.io import FilePointer
+from pyarrow import Schema
+import pyarrow.parquet as pq
```
delucchi-cmu, line 4: There are similar methods for both of these in `hipscat.io.file_io`. Are you planning to add more here that would necessitate having a separate io layer?

If we should move to the hipscat io methods, that can wait for a follow-up refactor.

src/lsdb/loaders/hipscat/__init__.py
delucchi-cmu, line 1: Why the additional directory nesting? What else will live under `loaders/`?

src/lsdb/loaders/hipscat/hipscat_catalog_loader.py
```diff
@@ -0,0 +1,98 @@
...
+import pyarrow
+from hipscat.pixel_math import HealpixPixel
+from hipscat.pixel_math.hipscat_id import HIPSCAT_ID_HEALPIX_ORDER
+
+from lsdb import io
+from lsdb.catalog.catalog import Catalog, DaskDFPixelMap
+from lsdb.loaders.hipscat.hipscat_loading_config import HipscatLoadingConfig
+
+
+# pylint: disable=R0903
```
delucchi-cmu, line 14: Can we either 1) use the human-readable pylint name or 2) adjust the min number of public methods in the .pylintrc? Again, something that can wait for a follow-up.

src/lsdb/loaders/hipscat/hipscat_loading_config.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class HipscatLoadingConfig:
```
delucchi-cmu, line 5: Why does this exist? What will be here eventually? Can you add that to the docstring?

src/lsdb/loaders/hipscat/read_hipscat.pyi
```diff
@@ -0,0 +1,14 @@
+from typing import overload, Type
```
delucchi-cmu, line 1: Can we get a module-level docstring to explain why we have this, and how it would be used?

tests/lsdb/catalog/test_catalog.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_catalog_compute_equals_ddf_compute(small_sky_order1_catalog):
+    pd.testing.assert_frame_equal(
+        small_sky_order1_catalog.compute(), small_sky_order1_catalog._ddf.compute()
+    )
+
+
+def test_get_catalog_partition_gets_correct_partition(small_sky_order1_catalog):
+    for _, row in small_sky_order1_catalog.hc_structure.get_pixels().iterrows():
```
delucchi-cmu, line 26: I see this too much and it grosses me out. Maybe the `hc.Catalog.get_pixels()` method should not be here, or should be replaced by `get_healpix_pixels` that returns a list of `HealpixPixel`. Yet again, something that can wait.

Though I think that the lsdb catalog should also offer a `get_healpix_pixels` convenience method (even if it's just wrapping the one on `hc.Catalog`).

delucchi-cmu (commented):
Mostly suggestions of things that you don't need to do now, but a few docstring requests.

src/lsdb/loaders/hipscat/__init__.py
smcguire-cmu: I think we've discussed in the past loading smaller files like CSVs directly, and I thought those could leave there. But it can be changed for now.

src/lsdb/io/parquet_io.py
smcguire-cmu: Yeah, this was written before hipscat had those methods, and I originally thought that hipscat wouldn't need to touch parquet files. But these should be replaced now they're in hipscat.

src/lsdb/loaders/hipscat/hipscat_catalog_loader.py
smcguire-cmu: Updated, thanks

src/lsdb/loaders/hipscat/hipscat_loading_config.py
smcguire-cmu: It's meant to contain whatever parameters are needed for loading a catalog. I figured there'd be quite a few of those and it'd be better to have an object to store them rather than passing around loads of individual variables. Originally I had a few in here, but then made some changes and realized they weren't necessary. I think as we develop more and use fsspec we'll probably need this so I kept it in.

src/lsdb/io/parquet_io.py
smcguire-cmu: Created an issue for this #20

tests/lsdb/catalog/test_catalog.py
smcguire-cmu: Made an issue for this #21

delucchi-cmu (approved):

PR #18, opened 2023-08-10
Update catalog loading to work with hipscat 0.1.1

smcguire-cmu:


src/lsdb/loaders/hipscat/hipscat_catalog_loader.py
```diff
@@ -1,18 +1,23 @@
...
 from lsdb.loaders.hipscat.hipscat_loading_config import HipscatLoadingConfig
 
 
 # pylint: disable=R0903
 class HipscatCatalogLoader:
     """Loads a HiPSCat formatted Catalog"""
 
-    def __init__(self, path: str, config: HipscatLoadingConfig) -> None:
+    def __init__(
+            self, path: str, config: HipscatLoadingConfig
```
delucchi-cmu, line 19: What's your formatter doing here? Is it maybe using the default 80 line limit?

src/lsdb/loaders/hipscat/read_hipscat.py
```diff
@@ -1,20 +1,35 @@
...
+) -> CatalogTypeVar | Catalog:
     """Load a catalog from a HiPSCat formatted catalog.
 
     Args:
         path: The path that locates the root of the HiPSCat catalog
-        source: Source to load the catalog from. Default is `None`, in which case the source is
-        inferred from the path. Currently supported options are:
-            -`'local'`: HiPSCat files stored locally on disk
+        catalog_type: Default `lsdb.Catalog` The type of catalog being loaded. Use by specifying the
+        lsdb class for that catalog.
```
delucchi-cmu, line 30:
```suggestion
             lsdb class for that catalog.
```

src/lsdb/loaders/hipscat/read_hipscat.py
```diff
@@ -28,6 +43,11 @@ def read_hipscat(
     }
     config = HipscatLoadingConfig(**config_args)
 
-    loader = HipscatCatalogLoader(path, config)
+    catalog_type_to_use = Catalog
```
delucchi-cmu, line 46: Why not just use a default arg value? or

```
if not catalog_type:
  catalog_type = Catalog
```

src/lsdb/loaders/hipscat/hipscat_loader_factory.py
```diff
@@ -0,0 +1,36 @@
...
+
+def get_loader_for_type(
+        catalog_type_to_use: CatalogTypeVar,
+        path: str,
+        config: HipscatLoadingConfig
+):
+    """Constructs a CatalogLoader that loads a Dataset of the specified type
+
+    Args:
+        catalog_type_to_use (Type[Dataset]): the type of catalog to be loaded. Uses the actual type
```
delucchi-cmu, line 25: Could we be generous, and accept those other types? Or is this a semi-protected method that will be called from within lsdb loaders?

src/lsdb/loaders/hipscat/read_hipscat.py
smcguire-cmu: It's to get the type checking to work. From what I can tell, I think because it's a TypeVar it doesn't like catalog_type being assigned to directly and gives errors, this was the only way I could get it to work. I did notice another Mypy error with this though so that's been updated now.

src/lsdb/loaders/hipscat/hipscat_catalog_loader.py
smcguire-cmu: I'm not sure, I'll have to check my config. I've changed it now though

src/lsdb/loaders/hipscat/hipscat_loader_factory.py
smcguire-cmu: Again it's for type checking reasons. I don't think there's a way to get it to understand the type of catalog it returns depends on the value of a string, so if users wanted it to type correctly they'd have to use casts. I think this is better, even If it might be slightly unusual to users who don't do type checking.

smcguire-cmu: Having said that, there might be a way to overload the method so that if it uses the type it'll return the right type for people who care about type checking, or just returns the base type if it's a string. Let me look into that.

smcguire-cmu: Ok so I got lost in a bit of a type checking rabbit hole, but I've updated the code so that we use this less now. This method is unchanged, but this is just a semi-protected method and shouldn't be used by an end user.

The `read_hipscat` method though will be used by the user and had the same problem. I've now changed it so that by default it will read the catalog_info file and use the type specified in there. The problem with this is the method is typed to return the base `Dataset` class. It will still actually return the correct type of object, but the type checker won't pick that up (They can be different because python). The `catalog_type` parameter is only used if a user wants to have proper type checking and so they can specify the correct type to make sure that happens.

I kinda don't like this, but I think it's the best option. The alternative is having different read methods for different types of catalogs, like `read_association_catalog`, `read_object_catalog`, `read_source_catalog`. It would make type checking a lot easier, but it feels kinda not very pythonic and a bit clunky, although I it is better from an OOP perspective. I think we've discussed this before and wanted the single read method, but I'm not sure which is better.

delucchi-cmu (approved):

PR #36, opened 2023-10-06
Load catalog from a Pandas Dataframe

### LINKED ISSUE #32: Add `from_pandas` method for loading a catalog from a pandas DataFrame
Generate a hipscat catalog from a pandas data frame by specifying the ra and dec columns and splitting the file to hipscat partitions.

Nice to have:
- row_limit per partition
  - Auto set row limit to make ~100mb partitions
- Option to use partition structure from another catalog

camposandro:
Loads LSDB catalog from data points specified in a single CSV file. Also fixes _pylint_ and _my_py_ issues that were triggering pipeline failures. Closes #32.


tests/lsdb/loaders/dataframe/test_read_dataframe.py
```diff
@@ -0,0 +1,40 @@
...
+    # Read CSV file for the small sky order 1 catalog
+    catalog = lsdb.read_dataframe(
+        path=small_sky_order1_csv,
+        catalog_name="small_sky_order1",
+        catalog_type=CatalogType.SOURCE
+    )
+    assert isinstance(catalog, lsdb.Catalog)
+
+    # Catalogs have the same information
+    assert catalog.hc_structure.catalog_info == small_sky_order1_catalog.hc_structure.catalog_info
```
smcguire-cmu, line 20: The test data shouldn't ever change, but if you're comparing the catalog_info against the small_sky_order1 catalog info, you could make it so the catalog_name and catalog_type you use to initialize come from the small_sky_order1 catalog

tests/lsdb/loaders/dataframe/test_read_dataframe.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # Data frames have the same data (column data types may differ)
+    pd.testing.assert_frame_equal(
+        catalog.compute().reset_index(drop=True),
+        small_sky_order1_catalog.compute().reset_index(drop=True),
+        check_column_type=False
+    )
+
+
```
smcguire-cmu, line 29: Some extra unit tests I would like:
 - check the partitions of the dask df match the partitions in the ddf_partition_map
 - check the partitions of the ddf_partition_map match the pixel_tree and partition_info in the hc_structure
 - check using non default ra + dec column names
 - only allow catalog types that use ra dec partitioning with data points

src/lsdb/loaders/dataframe/read_dataframe.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def read_dataframe(
+    path: str,
+    catalog_name: str = "",
+    catalog_type: CatalogType | None = None,
+    ra_column: str = "ra",
+    dec_column: str = "dec",
+    threshold: int = 50,
+) -> CatalogTypeVar | Dataset:
```
smcguire-cmu, line 17: There are a couple more catalog_info parameters, could add a **kwargs that's passed along for those.

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Args:
+            path (str): Path to a CSV file
+            threshold (int): The maximum number of data points per pixel
+            **kwargs: Arguments to pass to the creation of the catalog info
+        """
+        self._check_path_is_valid(get_file_pointer_from_path(path))
+        self.path = hc.io.get_file_pointer_from_path(path)
+        self.threshold = threshold
+        self.catalog_info = CatalogInfo(**kwargs)
```
smcguire-cmu, line 35: There are only certain `catalog_type` parameters that are ok here. E.g. you couldn't create an AssociationCatalog with this method. Could we add a check to see if it's a valid catalog type for this use case? "object" and "source" are the only valid options.

src/lsdb/loaders/dataframe/read_dataframe.py
smcguire-cmu, line 17: Woo typing. I'm guessing the return type here is copied from `read_hipscat` but that's a bit different. It can read any type of catalog, so it could return an `AssociationCatalog` for example. (It can't now, but that's because it's not finished.) But this just returns a `Catalog` so that should be the return type.

src/lsdb/loaders/dataframe/read_dataframe.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+
+from hipscat.catalog import CatalogType
+
+from lsdb.catalog.dataset.dataset import Dataset
+from lsdb.loaders.dataframe.dataframe_catalog_loader import DataframeCatalogLoader
+from lsdb.loaders.hipscat.hipscat_loader_factory import CatalogTypeVar
+
+
+def read_dataframe(
```
smcguire-cmu, line 10: Ah naming. So I think we're trying to match the pandas/dask convention for our loading methods. They use `read_...` for anything that's being read from disk, and `from_...` for anything that's being made from an object in memory. This is currently reading a csv file, but I think the desired version of this function is one that uses a Pandas Dataframe as an input, so that should be named `from_dataframe`

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -0,0 +1,160 @@
...
+            df,
+            highest_order=self.HISTOGRAM_ORDER,
+            ra_column=self.catalog_info.ra_column,
+            dec_column=self.catalog_info.dec_column,
+        )
+        return hc.pixel_math.compute_pixel_map(
+            raw_histogram, highest_order=self.HISTOGRAM_ORDER, threshold=self.threshold
+        )
+
+    def _load_dask_df_and_map(
```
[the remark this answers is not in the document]
camposandro: You're right, changed it to `_generate_dask_df_and_map`!

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Args:
+            df (pd.Dataframe): The catalog Pandas Dataframe
+            pixels (List[int]): The indices of the pixels inside the Healpix pixel
+
+        Returns:
+            The Pandas Dataframe containing the data points for the Healpix pixel
+        """
+        left_bound = healpix_to_hipscat_id(self.HISTOGRAM_ORDER, pixels[0])
+        right_bound = healpix_to_hipscat_id(self.HISTOGRAM_ORDER, pixels[-1] + 1)
```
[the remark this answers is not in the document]
camposandro: Sure, [here](https://github.com/astronomy-commons/lsdb/blob/116edca3619f2d17d4f993e2959faf88f74aec47/src/lsdb/loaders/dataframe/dataframe_catalog_loader.py#L157) it is! What do you think?

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -0,0 +1,160 @@
...
+        """Create the Dask Dataframe from the list of Healpix pixel Dataframes
+
+        Args:
+            pixel_dfs (List[pd.DataFrame]): The list of Healpix pixel Dataframes
+            schema (pd.Dataframe): The original Dataframe schema
+
+        Returns:
+            The catalog's Dask Dataframe
+        """
+        delayed_dfs = [delayed(pd.DataFrame)(df) for df in pixel_dfs]
```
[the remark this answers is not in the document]
camposandro: It actually does, nice catch!

src/lsdb/loaders/dataframe/read_dataframe.py
camposandro: Done! I've bundled all the _catalog_info_ arguments into kwargs.

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
camposandro: I've added a [method](https://github.com/astronomy-commons/lsdb/blob/116edca3619f2d17d4f993e2959faf88f74aec47/src/lsdb/loaders/dataframe/dataframe_catalog_loader.py#L38) that wraps the creation of the _catalog_info_ and raises an error in case the type isn't valid.

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -0,0 +1,160 @@
...
+        and assigns the hipscat_index column as the dataframe index.
+
+        Args:
+            df (pd.Dataframe): The catalog Pandas Dataframe
+        """
+        df["hipscat_index"] = compute_hipscat_id(
+            ra_values=df[self.catalog_info.ra_column],
+            dec_values=df[self.catalog_info.dec_column],
+        )
+        df.set_index("hipscat_index", inplace=True)
```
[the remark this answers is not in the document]
camposandro: Alright! For now I just fixed the name of the column and moved it to a class constant.

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Args:
+            pixel_dfs (List[pd.DataFrame]): The list of Healpix pixel Dataframes
+            schema (pd.Dataframe): The original Dataframe schema
+
+        Returns:
+            The catalog's Dask Dataframe
+        """
+        delayed_dfs = [delayed(pd.DataFrame)(df) for df in pixel_dfs]
+        return dd.from_delayed(delayed_dfs, meta=schema).repartition(self.PARTITION_SIZE)
```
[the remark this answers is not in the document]
camposandro: @smcguire-cmu I am now calculating the threshold based on the size of the loaded Dataframe ([calculate_threshold](https://github.com/astronomy-commons/lsdb/blob/c7f4d380cad723e9cd6b9d00d50650ee53cd86eb/src/lsdb/loaders/dataframe/from_dataframe.py#L39)), and one can specify the desired partition size using the respective argument on `from_dataframe`. 

What do you think of this change?

src/lsdb/loaders/dataframe/from_dataframe.py
```diff
@@ -0,0 +1,58 @@
...
+    Returns:
+        Catalog object loaded from the given parameters
+    """
+    if threshold is not None and partition_size is not None:
+        raise ValueError("Specify only one: threshold or partition_size")
+    if threshold is None and partition_size is None:
+        threshold = DataframeCatalogLoader.DEFAULT_THRESHOLD
+    elif threshold is None and partition_size is not None:
+        threshold = calculate_threshold(df, partition_size)
+    assert isinstance(threshold, int)
```
smcguire-cmu, line 38: I think this logic should be in the loader. So that if you want to use the loader elsewhere it'll be more feature complete

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -4,157 +4,178 @@
...
+from hipscat.catalog import CatalogType
 from hipscat.catalog.catalog_info import CatalogInfo
-from hipscat.io import FilePointer, get_file_pointer_from_path
 from hipscat.pixel_math import HealpixPixel, generate_histogram
 from hipscat.pixel_math.hipscat_id import compute_hipscat_id, healpix_to_hipscat_id
 
 from lsdb.catalog.catalog import Catalog, DaskDFPixelMap
-from lsdb.io.csv_io import read_csv_file_to_pandas
 
 HealpixInfo = NamedTuple("HealpixInfo", [("num_points", int), ("pixels", List[int])])
```
smcguire-cmu, line 14: Is it a NamedTuple?

smcguire-cmu: I think the `hipscat` method just returns an ordinary tuple, right?

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -4,157 +4,178 @@
...
         """Initializes the Hipscat Catalog object
 
         Args:
-            pixels (List[HealpixPixel]): The list of Healpix pixels
+            pixels (List[HealpixPixel]): The list of HEALPix pixels
 
         Returns:
             The Hipscat catalog object
         """
         return hc.catalog.Catalog(self.catalog_info, pixels)
```
smcguire-cmu, line 161: Why is this its own method?

tests/lsdb/loaders/dataframe/test_from_dataframe.py
```diff
@@ -0,0 +1,144 @@
...
+        lsdb.from_dataframe(small_sky_order1_df, **kwargs)
+    # If they were indeed named differently
+    small_sky_order1_df.rename(columns={"ra": "my_ra", "dec": "my_dec"}, inplace=True)
+    lsdb.from_dataframe(small_sky_order1_df, **kwargs)
+
+
+def test_partitions_obey_partition_size(small_sky_order1_df, small_sky_order1_catalog):
+    """Tests that partitions are limited by the specified size"""
+    partition_size_bytes = 1 << 10
+    partition_size_megabytes = partition_size_bytes / (1 << 20)
```
smcguire-cmu, line 109: Pretty sure this is equivalent to `partition_size_bytes >> 20`

tests/lsdb/loaders/dataframe/test_from_dataframe.py
```diff
@@ -0,0 +1,144 @@
...
+    with pytest.raises(KeyError):
+        lsdb.from_dataframe(small_sky_order1_df, **kwargs)
+    # If they were indeed named differently
+    small_sky_order1_df.rename(columns={"ra": "my_ra", "dec": "my_dec"}, inplace=True)
+    lsdb.from_dataframe(small_sky_order1_df, **kwargs)
+
+
+def test_partitions_obey_partition_size(small_sky_order1_df, small_sky_order1_catalog):
+    """Tests that partitions are limited by the specified size"""
+    partition_size_bytes = 1 << 10
```
smcguire-cmu, line 108: So this is because you want 1kb size partitions? Comment for this pls

tests/lsdb/loaders/dataframe/test_from_dataframe.py
```diff
@@ -0,0 +1,144 @@
...
+        lsdb.from_dataframe(small_sky_order1_df, **kwargs)
+
+
+def test_partitions_on_map_equal_partitions_in_df(small_sky_order1_df, small_sky_order1_catalog):
+    """Tests that partitions on the partition map exist in the Dask Dataframe"""
+    kwargs = get_catalog_kwargs(small_sky_order1_catalog)
+    catalog = lsdb.from_dataframe(small_sky_order1_df, **kwargs)
+    for _, partition_index in catalog._ddf_pixel_map.items():
+        partition_df = catalog._ddf.partitions[partition_index].compute()
+        assert isinstance(partition_df, pd.DataFrame)
```
smcguire-cmu, line 71: Can you check that each row in the df has ra dec within the pixel?

tests/lsdb/loaders/dataframe/test_from_dataframe.py
camposandro: We really need to perform the division because using right shift we would not get the partition size in megabytes as a float. In this case we are trying to partition by 1KB so that would happen.

smcguire-cmu (approved):

PR #37, opened 2023-10-09
Method to get HEALPix pixels from Catalog

### LINKED ISSUE #21: Add `get_healpix_pixels` method to Catalog
Currently to get the Healpix pixels in a catalog, the `hc_structure` must be used. This can be simplified by having a catalog method to access these more succinctly.

camposandro:
Adds utility method to the LSDB catalog to access HEALPix pixels from the underlying HiPSCat catalog. The method was named `get_pixels` instead of `get_healpix_pixels` to keep the convention already in use by HiPSCat. Closes #21.

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -5,6 +5,7 @@ from typing import Dict, Tuple, Type, cast
 
 import dask.dataframe as dd
 import hipscat as hc
+import pandas as pd
 from hipscat.pixel_math import HealpixPixel
 
 from lsdb.catalog.dataset.dataset import Dataset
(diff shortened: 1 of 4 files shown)


src/lsdb/catalog/catalog.py
```diff
@@ -47,6 +48,14 @@ def __init__(
         super().__init__(ddf, hc_structure)
         self._ddf_pixel_map = ddf_pixel_map
 
+    def get_pixels(self) -> pd.DataFrame:
```
smcguire-cmu, line 51: So in hipscat we have `get_pixels` and `get_healpix_pixels` which is a bit of a mess. I'd like both those methods available here, but I don't like that naming. I think it makes more sense to call this `get_partition_info` so we can use `get_pixels` to get the healpix pixels.

smcguire-cmu: If you wanted to also add `get_pixels` calling `get_healpix_pixels` in this pr, that would be amazing, but that can be left for another pr

smcguire-cmu: But then again `get_partition_info` sounds like it will get the partition info object. I kinda just hate the whole df.

smcguire-cmu: You know what, leave it, this isn't an issue for this pr, I'll make another issue for it.

smcguire-cmu (approved):

PR #40, opened 2023-10-16
Import HiPSCat index column

### LINKED ISSUE #38: Make a constant for hipscat_index column name
The name of the hipscat index column "_hipscat_index" is used as a literal across the code base in places. It should be a constant in the hipscat library and used from there.

camposandro:
Imports the "_hipscat_index" column from the HiPSCat library, instead of having it declared across the code base as a literal. It also temporarily installs HiPSCat from source to reduce the number of releases required. Closes #38.

Currently on hold waiting for merge of #36.


pyproject.toml
```diff
@@ -16,7 +16,6 @@ classifiers = [
 dynamic = ["version"]
 dependencies = [
     "dask",
-    "hipscat",
```
smcguire-cmu, line 19: Why's this here?

smcguire-cmu (approved):

PR #42, opened 2023-10-19
Re-factor get_pixels

delucchi-cmu:
In preparation for addressing https://github.com/astronomy-commons/hipscat/issues/142, this removes the `get_pixels` method in favor of a `get_healpix_pixels` method to mirror what will be available on the `Catalog` API.

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -48,13 +48,13 @@ class Catalog(Dataset):
         super().__init__(ddf, hc_structure)
         self._ddf_pixel_map = ddf_pixel_map
 
-    def get_pixels(self) -> pd.DataFrame:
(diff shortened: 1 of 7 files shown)


src/lsdb/catalog/catalog.py
```diff
@@ -48,13 +48,13 @@ def __init__(
         super().__init__(ddf, hc_structure)
         self._ddf_pixel_map = ddf_pixel_map
 
-    def get_pixels(self) -> pd.DataFrame:
+    def get_healpix_pixels(self) -> pd.DataFrame:
```
camposandro, line 51:
```suggestion
    def get_healpix_pixels(self) -> List[HealpixPixel]:
```

camposandro: Otherwise looks good!

camposandro (approved):

Issue #41, opened 2023-10-19
Re-implementation of kd-tree cross-matching


hombit:
**Problem**
The current `KdTreeCrossmatch` implementation uses gnomonic projection, which could be inaccurate for large healpix tiles, especially in the polar regions.

**Proposed solution**
I propose to use 3-D tree based on xyz coordinates on the unit sphere. It would:
1. Give the exact matches, because minimum 3-D distance corresponds to minimum great circle distance (minimum chord = minimum arc)
2. Distances given by the kd-tree could be directly used for the search radius filtering, we just need to convert search radius to the corresponding chord length

**Performance improvement opportunities**
The direct implementation of the proposed solution could have worse performance, because both kd-tree building and looking-up algorithms scale as ∼k. However, the proposed solution could apply few tricks to make it more performant:
- We can drop almost all the math associated with `_filter_indexes_to_threshold` and `_great_circle_dist`. The replacement of `frame_gnomonic` with spherical-to-euclidian coordinate transformation should have no-to-little performance impact.
- Practically, most of the cross-matches would have small radius and average number of matches to be less than `n_neighbors` (even if it is 1). That's why I believe that it would be beneficial to swap the query operations: first query for radius, and then filter by the first `n_neighbors`. With the current kd-tree realization this order would lead to even less accurate results.
- We can set the origin of the xyz coordinate system near the tile (for example rotating spherical system to put the tile to the north pole, and shifting the origin to x=0, y=0, z=1). It would allow us to use float32 dtype for the xyz positions and `KDTree` which would reduce memory usage

maxwest-uw:
since there are some trade-offs in speed and memory vs. accuracy, maybe it would be better to add a new cross-match routine and have the unit-sphere calculations be their own separate version, so that the user can weigh their computational needs themselves?

hombit:
I generally agree that having a choice is good. However I have many concerns about having two close but a bit different implementations:
1) Right now It is not clear to me if 3d-tree implementation would be (much) slower
2) Supporting two things is harder than supporting one thing
3) The most important one, I don't think that it is a good idea to give an inaccurate solution, which could lead to the results different from other implementations (including algorithms implemented not in LSDB). From the user's perspective it is not even clear what accuracy this solution has: large healpix tiles could lead to errors much beyond the single-float point accuracy. So even if we tell that this algorithm is faster, but less accurate, how do we make the user's choice to be informed about the level of accuracy?

maxwest-uw:
[...]

PR #43, opened 2023-10-24
3dtree cross-matching

hombit:
This implements #41, replacing the current kD-tree cross-matching algorithm `KdTreeCrossmatch` with an updated version using the Cartesian coordinates of the objects on the unit sphere instead of gnomonic projection. The gnomonic projection implementation is currently moved into new `KdTreeGnomonicCrossmatch` (few bugs fixed), but I plan to remove it in another PR.

The new implementation should be more accurate for the edge-cases, because 1) looking for neighbors in the Cartesian space is equivalent to spherical matches, 2) current implementation looking for neighbors in gnomonic projection coordinate space. I believe that the proposed method is equivalent to [one used by `astropy`](https://github.com/astropy/astropy/blob/7a502230364e57c98e6a5f0184d9afb8831e7dc4/astropy/coordinates/matching.py#L179).

The benchmarks shows that this implementation is also faster than the original one. Also the new code is more compact.

It adds `scipy` as a new explicit dependency, but the plan is to rid of `scikit-learn`, because I see no performance difference, while `scikit-learn` is a much heavier dependency which actually includes `scipy`.

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -2,17 +2,10 @@
 
 For more information on writing benchmarks:
 https://asv.readthedocs.io/en/stable/writing_benchmarks.html."""
-import math
 import os
 
-import healpy as hp
-import numpy as np
-import pandas as pd
-from hipscat.pixel_math.hipscat_id import HIPSCAT_ID_COLUMN
-from scipy.spatial import KDTree
-
 import lsdb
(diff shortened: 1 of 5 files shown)


smcguire-cmu:
Could we add [parameters](https://docs.pytest.org/en/7.1.x/example/parametrize.html) to the unit tests to get them to run with both the crossmatch implementations?

smcguire-cmu (approved):

PR #46, opened 2023-10-25
Wrappers for Dataframe query and assign operations

### LINKED ISSUE #45: Create wrappers to query and assign dataframe columns
It would be useful to have utility methods that perform query and assign operations on a catalog's data frame. 

The `query` method should wrap [Dask's dataframe query](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.query.html) and use a string expression to filter data by existing column values.

The `assign` method should wrap [Dask's data frame assign](https://docs.dask.org/en/latest/generated/dask.dataframe.DataFrame.assign.html) and allow for the use of callables, scalars or sequence-like values to create or overwrite columns on the existing data frame.

camposandro:
Creates wrappers for the `query` and `assign` operations on a catalog's data frame. Closes #45.

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -113,7 +113,7 @@ class Catalog(Dataset):
 
         Args:
             **kwargs: Arguments to pass to the assign method. This dictionary
-                should contain the column names as keys and either a lambda
+                should contain the column names as keys and either a
                 function or a 1-D Dask array as their corresponding value.
 
         Returns:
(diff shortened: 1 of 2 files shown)


src/lsdb/catalog/catalog.py
```diff
@@ -91,6 +91,37 @@ def name(self):
...
+            raise ValueError("Provided query expression is not valid")
+        ddf = self._ddf.query(expr)
+        return Catalog(ddf, self._ddf_pixel_map, self.hc_structure)
+
+    def assign(self, **kwargs) -> Catalog:
+        """Assigns new columns to a catalog
+
+        Args:
+            **kwargs: Arguments to pass to the assign method. This dictionary
+                should contain the column names as keys and either a lambda
```
smcguire-cmu, line 116:
```suggestion
                should contain the column names as keys and either a
```

smcguire-cmu (approved):

PR #61, opened 2023-11-06
Save catalog to disk

### LINKED ISSUE #50: Catalog saving
e.g. `to_hipscat` method.

It looks to be mostly implemented in this branch: https://github.com/astronomy-commons/lsdb/blob/sean/tape-demo-updated/src/lsdb/io/to_hipscat.py

camposandro:
Implements saving a LSDB catalog to disk in HiPSCat format. There is a [notebook](https://github.com/astronomy-commons/lsdb/blob/issue/50/catalog-saving/docs/notebooks/import_catalogs.ipynb) to compare the two possible scenarios we currently have to import a catalog (**lsdb.from_dataframe()** and the **hipscat-import** pipeline). Closes #50.


docs/notebooks/import_catalogs.ipynb
delucchi-cmu, line 1: We can address this in a follow-up PR, but this notebook isn't currently rendered with the rest of the documentation: https://lsdb--61.org.readthedocs.build/en/61/

src/lsdb/catalog/catalog.py
```diff
@@ -1,20 +1,31 @@
...
 from lsdb.core.crossmatch.crossmatch_algorithms import BuiltInCrossmatchAlgorithm
 from lsdb.dask.crossmatch_catalog_data import crossmatch_catalog_data
+from lsdb.io.file_io import write_dataframe_to_parquet
 
 DaskDFPixelMap = Dict[HealpixPixel, int]
 
+# Compute pixel map returns a tuple. The first element is
+# the number of data points within the HEALPix pixel, the
+# second element is the list of pixels it contains.
+HealpixInfo: TypeAlias = Tuple[int, List[int]]
```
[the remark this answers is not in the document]
delucchi-cmu: Or maybe not since I think this can be mostly removed when we take out `partition_info.csv`

src/lsdb/io/file_io.py
```diff
@@ -0,0 +1,21 @@
...
+    """Writes a pandas dataframe to parquet, creating the respective
+    parent directory if it does not yet exist
+
+    Args:
+        dataframe (pd.Dataframe): The pandas dataframe to write to disk
+        path (str): The path to write the parquet file to
+        storage_options (dict): dictionary that contains abstract filesystem credentials
+    """
+    parent_dir = os.path.dirname(os.path.abspath(path))
+    os.makedirs(parent_dir, exist_ok=True)
```
delucchi-cmu, line 20: Same as above - use the `hipscat.file_io` methods for anything where you're tempted to use `os` (except in unit tests - keep using `os` in unit tests).

src/lsdb/catalog/catalog.py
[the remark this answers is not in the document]
camposandro: Yes, I had moved it here because it was causing a circular dependency. In fact I think it doesn't make sense to have these types scattered all over the place so I moved HealpixInfo and DaskDFPixelMap to a self-contained module!

docs/notebooks/import_catalogs.ipynb
camposandro: I've fixed it :)

src/lsdb/io/to_hipscat.py
```diff
@@ -0,0 +1,149 @@
...
+        storage_options (dict): Dictionary that contains abstract filesystem credentials
+        **kwargs: Arguments to pass to the parquet write operations
+
+    Returns:
+        A dictionary mapping each HEALPix pixel to the number of data points in it.
+    """
+    pixel_to_partition_size_map = {}
+    for pixel, partition_index in catalog._ddf_pixel_map.items():
+        partition = catalog._ddf.partitions[partition_index].compute()
+        pixel_path = hc.io.paths.pixel_catalog_file(base_catalog_dir_fp, pixel.order, pixel.pixel)
```
delucchi-cmu, line 81: You can fetch the pixel directory path here with `pixel_directory`, instead of parsing it out of the full parquet file path later (https://github.com/astronomy-commons/hipscat/blob/main/src/hipscat/io/paths.py#L22).

This might be more straightforward as:

```
for pixel:
  io.make_directory(io.pixel_dir(base_dir, pixel))
  io.write_dataframe_to_parquet(...)
```

You wouldn't need the new `write_dataframe_to_parquet` method in this case.

src/lsdb/io/file_io.py
```diff
@@ -0,0 +1,23 @@
...
+    """Writes a pandas dataframe to parquet, creating the respective parent directory
+     if it does not yet exist
+
+    Args:
+        dataframe (pd.Dataframe): The pandas dataframe to write to disk
+        path (str): The path to write the parquet file to
+        storage_options (dict): dictionary that contains abstract filesystem credentials
+        **kwargs: Arguments to pass to the parquet write operation
+    """
+    parent_dir = os.path.dirname(os.path.abspath(path))
```
delucchi-cmu, line 20: We don't want any `os` calls (but if you implement another comment, then this is moot).

src/lsdb/io/to_hipscat.py
camposandro: I agree, thanks for the suggestion!

src/lsdb/io/file_io.py
camposandro: This method was removed according to the previous comment.

PR #65, opened 2023-11-14
Fix test data for removing partition_info

delucchi-cmu:
## Change Description

This is related to work toward deprecating `partition_info.csv` (https://github.com/astronomy-commons/hipscat/issues/147), and is very similar to PR https://github.com/astronomy-commons/hipscat-import/pull/166

## Solution Description

The catalog files for the cross matching test data is quite old, and didn't have the Norder/Dir/Npix columns inside the parquet file. When we try to read the directory as a hipscat catalog, we try to create the `PartitionInfo` from the row group statistics for those fields, but fail to do so.

This PR re-generates the catalog files, using the current map/reduce methods of the import tool to include the partition info columns as necessary.

While I'm here, I took out the tests for handling catalog directories with no metadata files (since we're going to require them very soon).


camposandro (approved):

camposandro:
There's the linting issues, but apart from that looks good! :)

PR #76, opened 2023-11-28
Fix index not being set correctly in `from_dataframe`

### LINKED ISSUE #75: `from_dataframe` doesn't set index column name
In the `DataframeCatalogLoader` class, when making the dask dataframe it uses a schema generated from the original dataframe's columns. This doesn't include the name for the index column, which causes the index column to be unnamed in the resulting dask data frame.

smcguire-cmu:
fixes #75


camposandro (approved):
Looks good - thanks for applying this fix!

Issue #68, opened 2023-11-14
Implement polygon search


camposandro:
Add the possibility to perform a polygonal search to filter points in a catalog. An initial implementation of this feature is present in this [branch](https://github.com/astronomy-commons/lsdb/blob/ADASS-Tutorial/src/lsdb/catalog/catalog.py). We need to productionize it and document it.

PR #77, opened 2023-11-29
Catalog search with polygonal regions

camposandro:
Implements polygonal searching for catalogs using astropy's [PolygonSkyRegion](https://astropy-regions.readthedocs.io/en/stable/api/regions.PolygonSkyRegion.html#regions.PolygonSkyRegion). This feature assembles a new catalog containing only the points within the polygonal region. It depends on the approval of hipscat's PR [!174](https://github.com/astronomy-commons/hipscat/pull/174). Closes https://github.com/astronomy-commons/lsdb/issues/68.


src/lsdb/core/search/polygon_search.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    data_frame["_POLYGON_PIX"] = hp.ang2pix(
+        2**max_order,
+        data_frame[metadata.catalog_info.ra_column].values,
+        data_frame[metadata.catalog_info.dec_column].values,
+        lonlat=True,
+        nest=True,
+    )
+    data_frame = data_frame.loc[data_frame["_POLYGON_PIX"].isin(polygon_pixels)]
+    data_frame = data_frame.drop(columns=["_POLYGON_PIX"])
```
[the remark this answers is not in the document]
camposandro: **For later reference:**

Following our discussions, we decided to use [spherical-geometry](https://github.com/spacetelescope/spherical_geometry). This package allows to define a `SingleSphericalPolygon` using vertices of (`ra`, `dec`) sky coordinates, and it contains helper methods that perform the needed spherical calculations to verify if the point is inside the polygon or not ([SingleSphericalPolygon.contains_lonlat](https://spherical-geometry.readthedocs.io/en/latest/api/spherical_geometry.polygon.SingleSphericalPolygon.html#spherical_geometry.polygon.SingleSphericalPolygon.contains_lonlat)). **astropy regions** seems to perform a projection of the polygon in the 2D plane and therefore would be another approach that follows an approximation.

src/lsdb/catalog/catalog.py
```diff
@@ -248,6 +251,31 @@ def cone_search(self, ra: float, dec: float, radius: float):
...
+        Args:
+            polygon (SingleSphericalPolygon): The polygon to filter pixels with.
+
+        Returns:
+            A new catalog containing the points filtered to those within the
+            polygonal region, and the partitions that have some overlap with it.
+        """
+        filtered_hc_structure = self.hc_structure.filter_by_polygon(polygon)
+        pixels_in_polygon = filtered_hc_structure.get_healpix_pixels()
+        partitions = self._ddf.to_delayed()
```
smcguire-cmu, line 269: There's actually now a `self.to_delayed()` that can be used here.

smcguire-cmu: If you copied this from the cone filter then I forgot to update it there, mb

src/lsdb/catalog/catalog.py
```diff
@@ -248,6 +251,31 @@ def cone_search(self, ra: float, dec: float, radius: float):
...
+        partitions_in_polygon = [partitions[self._ddf_pixel_map[hp_pixel]] for hp_pixel in pixels_in_polygon]
+        filtered_partitions = [
+            polygon_filter(partition, polygon, self.hc_structure) for partition in partitions_in_polygon
```
smcguire-cmu, line 272: So I hadn't actually thought about this before, but we're assuming the partition order of the filtered_hc_structure's partition_info is correct for the hipscat_index divisions here. The filtered_hc_structure is generate from an alignment, so it is in the correct order for now, but there might be issues down the line. I made #80 for this, so I don't think you have to do anything now.

src/lsdb/catalog/catalog.py
```diff
@@ -248,6 +251,31 @@ def cone_search(self, ra: float, dec: float, radius: float):
         ddf_partition_map = {pixel: i for i, pixel in enumerate(pixels_in_cone)}
         return Catalog(cone_search_ddf, ddf_partition_map, filtered_hc_structure)
 
+    def polygon_search(self, polygon: SingleSphericalPolygon) -> Catalog:
```
smcguire-cmu, line 254: Same comment about the polygon input types as with hipscat, could you make an issue for it here too?

tests/lsdb/catalog/test_polygon_search.py
```diff
@@ -0,0 +1,43 @@
...
+    polygon_search_catalog = small_sky_order1_catalog.polygon_search(polygon)
+    assert len(polygon_search_catalog.get_healpix_pixels()) == 0
+    assert len(polygon_search_catalog.hc_structure.pixel_tree) == 1
+
+
+def test_polygon_search_invalid_shape(small_sky_order1_catalog):
+    # Polygon is not convex, so the shape is invalid
+    ra, dec = [0, 1, 1, 0], [1, 0, 1, 0]
+    polygon = SingleSphericalPolygon.from_lonlat(ra, dec)
+    with pytest.raises(RuntimeError):
```
smcguire-cmu, line 42: This is a runtime error? Are there any more details in the error message?

tests/lsdb/catalog/test_polygon_search.py
```diff
@@ -0,0 +1,43 @@
+import pytest
+from spherical_geometry.polygon import SingleSphericalPolygon
+
+
+def test_polygon_search_filters_correct_points(small_sky_order1_catalog):
+    ra, dec = [300, 300, 272, 272], [-50, -55, -55, -50]
```
smcguire-cmu, line 6: What does this polygon cover? The entire catalog or just a part of it?

src/lsdb/catalog/catalog.py
camposandro: Got it, will update both!

src/lsdb/catalog/catalog.py
camposandro: Good point! I had not given a thought about divisions after filtering. I am not confident that the filtered catalogs keep any divisions anyway because again we create the catalog from scratch. We'll probably have to regenerate them.

src/lsdb/catalog/catalog.py
camposandro: Created the issue with a pointer to the one on hipscat (#81).

tests/lsdb/catalog/test_polygon_search.py
camposandro: The exception thrown is a `RuntimeError('Unknown exception')`, with no additional data. There is some output log stating that the "polygon is not convex" but it seems like it's being thrown in C code ([`healpix_base.cc/query_polygon_internal`](https://github.com/healpy/healpixmirror/blob/bc5cbaced09007fdda7f1fd9ea953d57dae7d2c5/src/cxx/Healpix_cxx/healpix_base.cc#L984C27-L984C49)) and not being sent up the chain.

src/lsdb/catalog/catalog.py
```diff
@@ -261,6 +263,31 @@ def cone_search(self, ra: float, dec: float, radius: float):
...
+            polygonal region, and the partitions that have some overlap with it.
+        """
+        filtered_hc_structure = self.hc_structure.filter_by_polygon(polygon)
+        pixels_in_polygon = filtered_hc_structure.get_healpix_pixels()
+        partitions = self.to_delayed()
+        partitions_in_polygon = [partitions[self._ddf_pixel_map[hp_pixel]] for hp_pixel in pixels_in_polygon]
+        filtered_partitions = [
+            polygon_filter(partition, polygon, self.hc_structure) for partition in partitions_in_polygon
+        ]
+        polygon_search_ddf = dd.from_delayed(filtered_partitions, meta=self._ddf._meta)
```
smcguire-cmu, line 286: I've left this so late that we can now add divisions in here.

camposandro: Nice catch! I also added them to the catalogs resulting from cone searches.

smcguire-cmu (approved):

PR #87, opened 2023-12-11
Extend specification of crossmatch extra columns

### LINKED ISSUE #66: Keep Dask "meta" in AbstractCrossmatchAlgorithm's children
Currently, `AbstractCrossmatchAlgorithm` has field `DISTANCE_COLUMN_NAME = "_DIST"`:
https://github.com/astronomy-commons/lsdb/blob/d57f77001dc19fce256f07ae7b747d5d6d544b8a/src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py#L12

This name is also hard-coded in `crossmatch_catalog_data()`:
https://github.com/astronomy-commons/lsdb/blob/d57f77001dc19fce256f07ae7b747d5d6d544b8a/src/lsdb/dask/crossmatch_catalog_data.py#L158

This is not very extendable: some cross-matching algorithms would like to have no distance output (Continuous maps) or multiple extra columns (`macauff`). I propose to replace `DISTANCE_COLUMN_NAME ` with `EXTRA_COLUMNS: pd.DataFrame` field, which would provide Dask meta for extra columns generated by `crossmatch()`.

camposandro:
Extends the ability to specify a set of crossmatch extra columns (in `pandas Series` format). It extends the behavior of the crossmatch algorithms which were previously limited to a distance output (`_DIST` column). Closes #66.


src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
```diff
@@ -9,7 +11,8 @@
 class AbstractCrossmatchAlgorithm(ABC):
     """Abstract class used to write a crossmatch algorithm"""
 
-    DISTANCE_COLUMN_NAME = "_DIST"
+    extra_columns: Dict[str, pd.Series] | None = None
```
hombit, line 14: 1. I believe that this `Union` syntax was introduced by Python 3.10
2. I think it would be better to use one of `meta` types supported by `dask` (see [this blog post](https://blog.dask.org/2022/08/09/understanding-meta-keyword-argument)), and do not introduce a new one. Could it be just `Dict[str, type]` or an empty `pd.DataFrame`?

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
```diff
@@ -58,3 +61,27 @@ def crossmatch(self) -> pd.DataFrame:
...
+
+    @classmethod
+    def _append_extra_columns(
+        cls, dataframe: pd.DataFrame, extra_columns: Dict[str, pd.Series] | None = None
+    ):
+        """Adds crossmatch extra columns to the resulting Dataframe."""
+        if cls.extra_columns is None:
+            return
+        if extra_columns is None:
+            raise ValueError("No extra column values were provided")
```
delucchi-cmu, line 73: @hombit can confirm here, but I think it's possible to not have any extra columns, e.g. in the case of SFD matching.  In that case, we're just adding in full row from the SFD catalog, and there's no meaningful distance or match likelihood.

Is the intent then that some `SFDCrossmatch` just never calls `_append_extra_columns`?

hombit (approved):
It looks good to me, but please consider to change the type of `extra_columns`

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
camposandro: Yes, that's exactly what I was thinking!

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
hombit: Thanks!

delucchi-cmu (approved):

PR #96, opened 2023-12-18
Write only non-empty partitions to disk

### LINKED ISSUE #93: `to_hipscat` doesn't work when one partition is empty
When running `gaia.cone_search(10,10,0.5).to_hipscat(...)` I get the following error:

```
RuntimeError: AppendRowGroups requires equal schemas.
The two columns with index 1 differ.
column descriptor = {
  name: designation,
  path: designation,
  physical_type: BYTE_ARRAY,
  converted_type: UTF8,
  logical_type: String,
  max_definition_level: 1,
  max_repetition_level: 0,
}
column descriptor = {
  name: designation,
  path: designation,
  physical_type: INT32,
  converted_type: NONE,
  logical_type: Null,
  max_definition_level: 1,
  max_repetition_level: 0,
}
```

with the end of the stack trace:

```
File ~/hipscat_deploy/src/hipscat/io/file_io/file_io.py:261, in write_parquet_metadata(schema, file_pointer, metadata_collector, storage_options, **kwargs)
    258 file_system, file_pointer = get_fs(file_pointer=file_pointer, storage_options=storage_options)
    260 file_pointer = strip_leading_slash_for_pyarrow(file_pointer, protocol=file_system.protocol)
--> 261 pq.write_metadata(
    262     schema, file_pointer, metadata_collector=metadata_collector, filesystem=file_system, **kwargs
    263 )

File ~/anaconda3/envs/lsdb/lib/python3.9/site-packages/pyarrow/parquet/core.py:3555, in write_metadata(schema, where, metadata_collector, filesystem, **kwargs)
   3552     where.seek(cursor_position)  # file-like, set cursor back.
   3554 for m in metadata_collector:
-> 3555     metadata.append_row_groups(m)
   3556 if filesystem is not None:
   3557     with filesystem.open_output_stream(where) as f:

File ~/anaconda3/envs/lsdb/lib/python3.9/site-packages/pyarrow/_parquet.pyx:799, in pyarrow._parquet.FileMetaData.append_row_groups()
```

I believe this is because the metadata collector gets different types for the different files because one is empty.

camposandro:
When writing a catalog to disk remove its empty partitions because the `_metadata` file does not have its corresponding statistics. Closes #93.


src/lsdb/io/to_hipscat.py
```diff
@@ -119,11 +119,25 @@ def write_partitions(
     pixel_to_result_index = {}
 
     partitions = catalog._ddf.to_delayed()
+
+    # Get the indices of non-empty partitions
+    partition_sizes = catalog._ddf.map_partitions(len).compute()
```
smcguire-cmu, line 124: So this works, but it needs the entire catalog to be computed twice (once to get the length of each partition, and one to actually write the results. Could we instead move this length check inside the `perform_write` method?

Then when we compute the partition_sizes, we can perform the `len(non_empty_partitions)` check there.

src/lsdb/io/to_hipscat.py
```diff
@@ -119,16 +121,28 @@ def write_partitions(
...
+                storage_options,
+                **kwargs,
+            )
         )
         pixel_to_result_index[pixel] = index
 
     partition_sizes = dask.compute(*results)
 
+    if all(size == 0 for size in partition_sizes):
+        raise AssertionError("The output catalog is empty")
```
smcguire-cmu, line 140: Can this be changed from AssertionError? Maybe something a runtime error instead?

camposandro: Sure, just pushed the changes!

smcguire-cmu (approved):

PR #101, opened 2023-12-21
Migrate to the `sphgeom` library for polygonal searches

### LINKED ISSUE #81: Allow input coordinates in the polygon search call
The interface should allow for a [polygon search](https://github.com/astronomy-commons/lsdb/blob/ec66a788dbf16702cf5b7d653a39d02c212aeedc/src/lsdb/catalog/catalog.py#L254C7-L254C7) using sky coordinates on the call directly. Related to issue [#177](https://github.com/astronomy-commons/hipscat/issues/177) on hipscat.

### LINKED ISSUE #84: Consider sphgeom package for vectorized region search
The LSST DM team has been developing a spherical geometry package ([sphgeom](https://github.com/lsst/sphgeom)) to use with Rubin. It would be worth assessing its performance in comparison with the one we are currently using (`spherical-geometry`), in particular because `region.contains()` seems to be vectorized.

camposandro:
It replaces the usage of `spherical-geometry` with `sphgeom` to perform the polygonal sky filtering in a vectorized way. It also changes the interface to accept as input from the user the lists of vertices instead of a pre-instantiated polygon object. Closes #81 and closes #84.


requirements.txt
```diff
@@ -1 +1 @@
-git+https://github.com/astronomy-commons/hipscat.git@main
\ No newline at end of file
+git+https://github.com/astronomy-commons/hipscat.git@issue/84/add-sphgeom
```
camposandro, line 1: After review, and before merging, I will revert this change. Same applies for `docs/requirements.txt`.

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "deprecated",
     "ipykernel", # Support for Jupyter notebooks
     "scipy", # kdtree
+    "lsst-sphgeom", # To handle spherical sky polygons
```
delucchi-cmu, line 24: Is this package conda-installable? No worries this moment if not, but it's good to know.

src/lsdb/core/search/polygon_search.py
```diff
@@ -1,28 +1,48 @@
...
+import numpy as np
 import pandas as pd
-from spherical_geometry.polygon import SingleSphericalPolygon
+from hipscat.pixel_math.polygon_filter import CartesianCoordinates, SphericalCoordinates
+from lsst.sphgeom import ConvexPolygon, UnitVector3d
 
 
 @dask.delayed
-def polygon_filter(data_frame: pd.DataFrame, polygon: SingleSphericalPolygon, metadata: hc.catalog.Catalog):
+def polygon_filter(data_frame: pd.DataFrame, polygon: ConvexPolygon, metadata: hc.catalog.Catalog):
```
delucchi-cmu, line 13: How "heavy" is the `ConvexPolygon` object? Will it be a problem to hand the whole object between dask workers?

tests/conftest.py
```diff
@@ -111,16 +111,18 @@ def xmatch_correct_3n_2t_no_margin(small_sky_xmatch_dir):
...
-    # Check that the divisions are not None
-    assert None not in catalog._ddf.divisions
-    # Check that divisions belong to the correct pixel
-    for division, hp_pixel in zip(catalog._ddf.divisions, hp_pixels):
-        div_pixel = hipscat_id_to_healpix([division], target_order=hp_pixel.order)
-        assert hp_pixel.pixel == div_pixel
-    # The last division corresponds to the HIPSCAT_ID_MAX
-    assert catalog._ddf.divisions[-1] == HIPSCAT_ID_MAX
+@pytest.fixture
+def assert_divisions_are_correct():
```
delucchi-cmu, line 115: =[

I'm sorry you had to do this.

src/lsdb/catalog/catalog.py
```diff
@@ -221,20 +221,22 @@ def cone_search(self, ra: float, dec: float, radius: float):
...
-    def polygon_search(self, polygon: SingleSphericalPolygon) -> Catalog:
+    def polygon_search(self, vertices: List[SphericalCoordinates]) -> Catalog:
         """Perform a polygonal search to filter the catalog.
 
         Filters to points within the polygonal region specified in ra and dec, in degrees.
         Filters partitions in the catalog to those that have some overlap with the region.
 
         Args:
-            polygon (SingleSphericalPolygon): The polygon to filter pixels with.
+            vertices (List[SphericalCoordinates]): The list of vertices of the polygon to
```
smcguire-cmu, line 231: I think the type alias in the docstring isn't as helpful as writing out the full tuple type. It's nice in the code, but I think for a user there's no way of finding out what the type should be without digging into the code and inspecting the definition in hipscat.

pyproject.toml
camposandro: Unfortunately it isn't, but considering @mjuric has taken the initiative of creating the recipe to make lsdb conda installable (https://github.com/conda-forge/staged-recipes/pull/24833) I might try to do the same for sphgeom.

tests/conftest.py
camposandro: I know right! It works but the linter was complaining about it at a certain point :((

src/lsdb/catalog/catalog.py
camposandro: I agree, from the users' perspective having the explicit type makes their life easier. I'm changing it!

src/lsdb/core/search/polygon_search.py
camposandro: Good question! It's hard to be sure but I think it shouldn't be a problem because this object is fully implemented in C++ under the hood and it only references the region class. I did some digging on how much an example polygon object with 4 to 6 vertices was weighing in memory on my machine and it seems to be under a steady 100 bytes.

delucchi-cmu (approved):

PR #117, opened 2024-01-10
Validate input coordinates for cone / polygonal searches

### LINKED ISSUE #100: cone_search() should accept RA>=180
Currently `Catalog.cone_search()` raises an error if `ra >= 180`, which should not happen since RA is usually considered to be in range of [0; 360). It would be nice to support any RA values, maybe implementing something like astropy's [`Angle.wrap_at`](https://docs.astropy.org/en/stable/api/astropy.coordinates.Angle.html#astropy.coordinates.Angle.wrap_at).

camposandro:
Removes the previous restriction of having right ascension defined in the [-180,180] degree range, wrapping the values provided by the user to [0,360]. This applies to both types of search (cone and polygonal). Closes #100.


src/lsdb/core/search/validators.py
delucchi-cmu, line 1: This looks like a nice encapsulation. Should it be in the hipscat package? What happens now if we use invalid coordinates in the cone or polygon search on the `hipscat.catalog.Catalog`?

src/lsdb/core/search/validators.py
```diff
@@ -0,0 +1,68 @@
...
+            ra (float): The right ascension to the center of the cone
+            dec (float): The declination to the center of the cone
+            radius (float): The cone radius, in degrees
+
+        Returns:
+            The wrapped value for the right ascension of the cone center, in the
+            [0,360] degrees range.
+        """
+        if radius < 0:
+            raise ValueError("Cone radius must be non negative")
```
delucchi-cmu, line 28: Why is a radius of 0 a valid value?

tests/lsdb/catalog/test_validators.py
```diff
@@ -0,0 +1,60 @@
...
+import numpy.testing as npt
+import pytest
+
+from lsdb.core.search.polygon_search import get_cartesian_polygon
+from lsdb.core.search.validators import CoordinatesValidator
+
+
+def test_wrap_right_ascension():
+    # Does not need to be wrapped
+    ra = [20, 150, 350]
```
delucchi-cmu, line 11: Can you add some test cases that use floats (e.g. `23.4`, `-89.999`)? Modular division and floats make me nervous.

src/lsdb/core/search/validators.py
camposandro: In reality, now that I look at it again, we don't even need to wrap the right ascension values for the searches. In both cases healpy's `ang2vec` seems to already do it by default whenever we use degrees instead of radians. 

We are left with validating the declination values. On the hipscat catalog a declination outside of [-90,90] / negative radius throws an error, maybe it's best if we port this validation over.

src/lsdb/core/search/validators.py
camposandro: Good question, it's maybe a typo! @hombit @smcguire-cmu do you know if there's any reason for this?

camposandro: I changed it so now radii need to be positive. Please let me know if there's an issue with it!

tests/lsdb/catalog/test_validators.py
delucchi-cmu: This still makes me nervous.

tests/lsdb/catalog/test_polygon_search.py
```diff
@@ -56,23 +56,29 @@ def test_polygon_search_invalid_shape(small_sky_order1_catalog):
...
-        - [(340, 1), (340, -1), (20, -1), (20, 1)]
-        - [(340, 1), (340, -1), (-340, -1), (-340, 1)]
-    """
+    polygon we have several possible combination of coordinates
+    (here with some float-fudging)."""
+    def get_vertex_combinations(vertex):
+        ra_combinations = (vertex[0], vertex[0] + 360, vertex[0] - 360)
+        return [(ra, vertex[1]) for ra in ra_combinations]
+
+    def calculate_polygon_combinations(vertices):
```
delucchi-cmu, line 65: This is more thorough, but I think the explicit list makes it easier to understand what's being tested (and I can kind of reason about the math when I see some real numbers).

camposandro: Thanks, makes sense to me - just added the explicit values but reduced the polygon to a triangle so that we have fewer combinations to copy-paste!

delucchi-cmu (approved):

Issue #121, opened 2024-01-16
Missing jinja2 and dask[diagnostics] packages when running lsdb.read_hipscat()


stargaser:
I am running a notebook in which I'm using lsdb 0.1 (installed with pip into a Python 3.11.7 conda environment). Upon using `lsdb.read_hipscat()` I get an error `ModuleNotFoundError: No module named 'jinja2'`.

Then the traceback later says

```
The above exception was the direct cause of the following exception:
```

followed by 

```
ImportError: Dask diagnostics requirements are not installed.

Please either conda or pip install as follows:

  conda install dask                     # either conda install
  python -m pip install "dask[diagnostics]" --upgrade  # or python -m pip install
```

Shouldn't a `pip install lsdb` have installed these for me?

stargaser:
Later in my notebook I got another error:

```
ImportError: dask.distributed is not installed.

Please either conda or pip install distributed:

  conda install dask distributed             # either conda install
  python -m pip install "dask[distributed]" --upgrade    # or pip install
```

Same question, shouldn't this have been installed automatically?

delucchi-cmu:
@hombit has tried to re-create this issue unsuccessfully. Can you please attach the notebook you're working with that causes this issue?

stargaser:
I've attached the executed notebook (it's in MyST Markdown format so needs jupytext) and the requirements.txt file as it now stands, with the extra items added in.

Initially I created a conda environment with python=3.11 as detailed in the "Environment Setup" section of the notebook.

[requirements.txt](https://github.com/astronomy-commons/lsdb/files/13966172/requirements.txt)
[CosmoDC2_HiPSCat.md](https://github.com/astronomy-commons/lsdb/files/13966171/CosmoDC2_HiPSCat.md)

stargaser:
The catalog accessed by the notebook was generated using hipscat-import v0.1. In a few weeks, @troyraen will regenerate it with hipscat-import v0.2 and then we plan to put it in the cloud.

mjuric:
In case it's related, running `pytest` on lsdb with only the runtime requirements triggers the missing `jinja2 error.

```
$PREFIX/lib/python3.11/site-packages/dask/widgets/__init__.py:4:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    from __future__ import annotations

    import datetime
    import html
[...]

PR #122, opened 2024-01-17
Clean up some dependencies

delucchi-cmu:
This is related to #121 , but doesn't really close it.

- we're using `dask[distributed]` and `dask[dataframe]`, so let's explicitly depend on them.
- remove extra spherical geometry dep on docs
- maybe fix github being sad in `pyproject.toml` about `project.urls`?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,7 @@
 name = "lsdb"
 license = {file = "LICENSE"}
 readme = "README.md"
+description = "Spatial analysis for extremely large astronomical databases using dask"
 authors = [
     { name = "LINCC Frameworks", email = "lincc-frameworks-team@lists.lsst.org" }
 ]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -16,16 +17,18 @@ classifiers = [
 dynamic = ["version"]
 dependencies = [
     "dask",
+    "dask[dataframe]",
+    "dask[distributed]",
```
hombit, line 21: I believe that `dask[distributed]` is not a strict requirement, but something we'd like user to have to be able to run things in parallel with Dask client. User may still run LSDB without Dask client, 1) with no client and not in parallel, 2) with [Dask on Ray](https://docs.ray.io/en/latest/ray-more-libs/dask-on-ray.html), 3) maybe something else.

I would make a comment here about it.

hombit, line 21: I believe these three lines should be collapsed into `dask[dataframe,distributed]`

hombit (changes requested):
Thank you! I request a minor change of the dependency list

hombit (approved):
LGTM!

PR #124, opened 2024-01-17
Add more thorough polygon validation

### LINKED ISSUE #120: Check for polygon degeneracy in search
Healpy errors (from its internal C code) when we try to perform a search with a degenerate polygon. 

This behavior is expected and we can perceive it as a runtime exception up the chain in our Python program. However, unfortunately, it carries no relevant information so it's not possible to understand what was type of exception being thrown. By validating the degeneracy of the polygon ourselves we make our API sturdier.

We will only consider polygons with 3+ vertices. In this scenario, the polygon is degenerate when all the points fall in a great circle:

camposandro:
Validates polygons used for catalog search. They must contain at least three vertices (without duplicates) and form a non-degenerate region (i.e., the vertices cannot be on a single great circle). Closes #120.

### DIFF
--- a/src/lsdb/core/search/polygon_search.py
+++ b/src/lsdb/core/search/polygon_search.py
@@ -11,7 +11,7 @@ from hipscat.pixel_math.polygon_filter import (
     SphericalCoordinates,
     filter_pixels_by_polygon,
 )
-from hipscat.pixel_math.validators import validate_declination_values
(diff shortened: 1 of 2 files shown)


camposandro:
The checks will fail until https://github.com/astronomy-commons/hipscat/pull/195 is approved.

PR #134, opened 2024-01-27
Change default crossmatch radius

### LINKED ISSUE #128: Change default crossmatching radius to 1 arcsec
Change default crossmatching radius to 1 arcsec (from very large value of 0.01 deg)

smcguire-cmu:
Changes default crossmatch maximum distance from 0.01 degrees, to 1 arc second

closes #128

### DIFF
--- a/src/lsdb/core/crossmatch/kdtree_match.py
+++ b/src/lsdb/core/crossmatch/kdtree_match.py
@@ -18,7 +18,7 @@ class KdTreeCrossmatch(AbstractCrossmatchAlgorithm):
     def crossmatch(
         self,
         n_neighbors: int = 1,
-        d_thresh: float = 0.000278,  # 1 arcsecond
+        radius_arcsec: float = 1,  # 1 arcsecond
     ) -> pd.DataFrame:
         """Perform a cross-match between the data from two HEALPix pixels
 
@@ -27,7 +27,7 @@ class KdTreeCrossmatch(AbstractCrossmatchAlgorithm):
 
         Args:
             n_neighbors (int): The number of neighbors to find within each point
-            d_thresh (float): The threshold distance in degrees beyond which neighbors are not added
+            radius_arcsec (float): The threshold distance in arcseconds beyond which neighbors are not added
 
         Returns:
             A DataFrame from the left and right tables merged with one row for each pair of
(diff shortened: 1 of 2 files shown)


src/lsdb/core/crossmatch/kdtree_match.py
```diff
@@ -18,7 +18,7 @@ class KdTreeCrossmatch(AbstractCrossmatchAlgorithm):
     def crossmatch(
         self,
         n_neighbors: int = 1,
-        d_thresh: float = 0.01,
+        d_thresh: float = 0.000278,
```
camposandro, line 21: Can we add a comment saying this represents 1 arc second, to make it more explicit?

camposandro (commented):
We should regenerate the test crossmatch results because they depend on this threshold value

camposandro (approved):

src/lsdb/core/crossmatch/kdtree_match.py
```diff
@@ -18,7 +18,7 @@ class KdTreeCrossmatch(AbstractCrossmatchAlgorithm):
     def crossmatch(
         self,
         n_neighbors: int = 1,
-        d_thresh: float = 0.01,
+        d_thresh: float = 0.000278,  # 1 arcsecond
```
delucchi-cmu, line 21: Would it make sense to change the units on this input? If the values are generally going to be less than a degree, then it might be silly, but if astronomers expect degrees then it's probably ok. Really, I'm asking @hombit or @nevencaplar though.

hombit: 1. I would write `1.0 / 3600.0` instead here (if we keep degrees)
2. There is no single value people use across the APIs and libraries, I've seen all possible units: degrees, arcminutes, arcseconds, and even radians. I would prefer to use arcseconds, because it the typical scale we are aimed to support
3. I would rename the argument to something like `radius_arcsec` to make the units more explicit for users

nevencaplar: I support arcseconds. @delucchi-cmu Are there many place where we would need to change API? If yes, does that deserve another ticket?

smcguire-cmu: Does everyone agree on the `radius_arcsec` name change? Should we have more explicit units in names everywhere? The only other radius I can think of right now is the cone search radius. Does that make more sense in arcsec too?

delucchi-cmu (approved):

hombit (approved):

PR #138, opened 2024-01-31
Add catalog box searching

### LINKED ISSUE #90: Functionality of read_hipscat to import only part of the data
At the moment, it seems that it is only possible to load the full catalog at one time (e.g., 

`lsdb.read_hipscat("/epyc/data3/hipscat/catalogs/ztf_axs/ztf_source")`

). That is very reasonable, but can be very taxing for the user, especially if one only wants to do some test of the algorithm, or if the kernel fails.

It would be useful if there would be a way to import a subpart of the catalog. In ideal circumstances, instead of reading the entire catalog as above, I can image something like: 
`lsdb.read_hipscat("/epyc/data3/hipscat/catalogs/ztf_axs/ztf_source", ra=[10,20], dec[7,23])` giving the user parts of the catalog that contains data from those parts of the sky?

In this way, the user could test their algorithm and procedures on a subset of the real data, before potentially scaling up their algorithm on the full sky investigation.

camposandro:
Adds box searching for catalogs.

### Solution description

This type of search uses ranges of right ascension (ra) and/or declination (dec) to define a subregion of the sky. It obtains the corresponding partitions and filters their data points so that we only keep those that are inside the defined subregion. If both ra and dec are provided we construct a spherical polygon and use the preexisting polygon search.

Closes #90. Depends on https://github.com/astronomy-commons/hipscat/pull/202.

### DIFF
--- a/src/lsdb/core/search/box_search.py
+++ b/src/lsdb/core/search/box_search.py
@@ -1,0 +1,73 @@
+from __future__ import annotations
+
+from typing import List, Tuple
+
+import hipscat as hc
+import numpy as np
+import pandas as pd
+from hipscat.pixel_math import HealpixPixel
+from hipscat.pixel_math.box_filter import filter_pixels_by_box, transform_radec, wrap_angles
+from hipscat.pixel_math.validators import validate_box_search
+from hipscat.pixel_tree.pixel_tree_builder import PixelTreeBuilder
+
+from lsdb.core.search.abstract_search import AbstractSearch
+
+
+class BoxSearch(AbstractSearch):
+    """Perform a box search to filter the catalog. This type of search is used for a
+    range of ra or dec. If both, a polygonal search is the better option.
+
+    Filters to points within the ra / dec region, specified in degrees.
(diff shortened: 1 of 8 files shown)


src/lsdb/core/search/box_search.py
```diff
@@ -0,0 +1,89 @@
...
+    ra: Tuple[float, float] | None,
+    dec: Tuple[float, float] | None,
+    metadata: hc.catalog.Catalog,
+):
+    """Filters a dataframe to only include points within the specified box region.
+
+    Args:
+        data_frame (pd.DataFrame): DataFrame containing points in the sky
+        ra (Tuple[float, float]): Right ascension range, in degrees
+        dec (Tuple[float, float]): Declination ascension range, in degrees
```
delucchi-cmu, line 59:
```suggestion
        dec (Tuple[float, float]): Declination range, in degrees
```

delucchi-cmu (approved):

smcguire-cmu (approved):

PR #144, opened 2024-02-06
Add margin to catalog and use in crossmatch

smcguire-cmu:
Adds `margin` as a parameter in the `Catalog` class and the `read_hipscat` method. Changes crossmatch to use it.


src/lsdb/dask/merge_catalog_functions.py
```diff
@@ -18,15 +18,52 @@
...
+            right.hc_structure.pixel_tree,
+            right.margin.hc_structure.pixel_tree,
+            alignment_type=PixelAlignmentType.OUTER,
+        ).pixel_tree
+    else:
+        right_tree = right.hc_structure.pixel_tree
+    return align_trees(left.hc_structure.pixel_tree, right_tree, alignment_type=PixelAlignmentType.INNER)
+
+
+def join_partition_and_margin(df: pd.DataFrame, margin_df: pd.DataFrame, columns: List[str]) -> pd.DataFrame:
```
delucchi-cmu, line 42: This feels like a bit of an extreme to extract into a helper method. Do you anticipate that this would be called in many places? Could it just be inline instead?

delucchi-cmu (approved):
One small comment. This is happening!!!

src/lsdb/dask/merge_catalog_functions.py
smcguire-cmu: Yeah, that's fair. It'll be used in the join methods, but it doesn't really need a method. I'll inline it

PR #150, opened 2024-02-08
Add `head` method to catalog

### LINKED ISSUE #106: head() operation to fetch couple of rows from a single partition to preview the data
Make `head` operation to quickly look at the data avaliable at `lsdb.Catalog` level. 

For instance, if we load `gaia`, at the moment to look at the data one has to do something like `gaia._ddf.head()`. The goal is to make `head` more user-friendly so that we dont have to go via `_ddf` to look at the data.

camposandro:
Adds a simple `head` method to the catalog to get a few rows of data for previewing purposes. Closes #106.

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -76,7 +76,15 @@ class Catalog(HealpixDataset):
         Returns:
             A pandas DataFrame with up to `n` of data.
         """
-        return self._ddf.head(n, npartitions=-1)
+        dfs = []
+        remaining_rows = n
+        for partition in self._ddf.partitions:
+            if remaining_rows == 0:
+                break
+            partition_head = partition.head(remaining_rows)
+            dfs.append(partition_head)
+            remaining_rows -= len(partition_head)
+        return pd.concat(dfs)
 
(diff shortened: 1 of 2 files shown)


src/lsdb/catalog/catalog.py
```diff
@@ -66,6 +67,17 @@ def get_ordered_healpix_pixels(self) -> List[HealpixPixel]:
...
+    def head(self, n: int = 5) -> pd.DataFrame:
+        """Returns a few rows of data for previewing purposes.
+
+        Args:
+            n (int): The number of desired rows.
+
+        Returns:
+            A pandas DataFrame with up to `n` of data.
+        """
+        return self._ddf.head(n, npartitions=-1)
```
hombit, line 79: It looks like `npartitions=-1` makes Dask loading all partitions, which is very slow. We can go smarter and optimize it for small `n`, sequentially examining partitions until we collect `n` rows.

camposandro: @hombit thanks for the feedback, you're right! I've applied a fix.

delucchi-cmu (approved):

hombit (approved):

PR #154, opened 2024-02-09
Docs: fix link to notebooks page

hombit:


docs/index.rst
```diff
@@ -17,7 +17,7 @@ You can install LSDB with `conda` or `pip`:
    >> # Pip installation
    >> python3 -mpip install lsdb
 
-See `notebooks <notebooks>`_ for examples.
+See `notebooks <notebooks.html>`_ for examples.
```
drewoldag, line 20: It's odd that this is required for the link, but as long as it's working I won't stand in the way.

drewoldag (approved):
Looks ok. Seems odd that this change is required, but if it works it works.

PR #157, opened 2024-02-13
Fix inconsistent `_hipscat_index` type

### LINKED ISSUE #145: Inconsistent dtype for _hipscat_index (int64 vs uint64)
In https://github.com/lincc-frameworks/tape/issues/373, I observed in a run of the [Full FG Dwarf Dipper Search notebook](https://github.com/dirac-institute/ZTF_FG_BoyajianSearch/blob/main/analysis/notebooks/paper/Full%20FG%20Dwarf%20Dipper%20Search.ipynb) inconsistent behavior over whether the `_hipscat_index` column has a dtype of `int64` or `uint64`.

Further inspection within the notebook suggests that this can be the result of joining two lsdb catalogs. See the following screenshots:

image

image

Is this behavior expected? We are joining on columns which are not the hipscat index, but the resulting catalog still uses `_hipscat_index` as its dask dataframe index, and I would expect the dtype to still remain a `uint64`.

camposandro:
Forces the `_hipscat_index` type on joined catalogs to be `uint64`. We were not enforcing the index type before so Pandas was infering `int64`, which is incorrect. Closes #145.

### DIFF
--- a/src/lsdb/dask/merge_catalog_functions.py
+++ b/src/lsdb/dask/merge_catalog_functions.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Callable, List, Sequence, Tuple, cast
 
 import dask.dataframe as dd
 import numpy as np
+import numpy.typing as npt
 import pandas as pd
 from dask.delayed import Delayed
 from hipscat.pixel_math import HealpixPixel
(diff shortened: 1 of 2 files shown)


wilsonbb (approved):
Thanks Sandro!

delucchi-cmu (approved):

Issue #137, opened 2024-01-30
Address warnings during pytest


delucchi-cmu:
The following warnings are raised during test execution:

```
=============================== warnings summary ===============================
tests/lsdb/catalog/test_index_search.py::test_index_search
  /opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/hipscat/pixel_math/healpix_pixel_function.py:21: DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
    if not pixels:

tests/lsdb/catalog/test_merge.py::test_catalog_merge_on_indices[left]
tests/lsdb/catalog/test_merge.py::test_catalog_merge_on_indices[left]
  /opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/partd/pandas.py:6: DeprecationWarning: create_block_manager_from_blocks is deprecated and will be removed in a future version. Use public APIs instead.
    from pandas.core.internals import create_block_manager_from_blocks, make_block

tests/lsdb/catalog/test_merge.py: 212 warnings
  /opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/pandas/core/frame.py:706: DeprecationWarning: Passing a BlockManager to DataFrame is deprecated and will raise in a future version. Use public APIs instead.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================= 77 passed, 215 warnings in 5.63s =======================
```

delucchi-cmu:
There are just a few remaining warnings
- block manager warnings are coming from dask's use of pandas, and will be addressed in future version
- dask dataframe deprecation warning tracked in #181

PR #168, opened 2024-02-19
Catch warnings, and add data to avoid them.

delucchi-cmu:
Related to #137. This does not address new deprecation warnings.


camposandro (approved):
Awesome, thanks! 

By the way, I noticed the deprecation warnings issued in CI started appearing after the most recent releases of Dask (v2024.2.0) and pandas (v2.2.0). Hope this is useful to know!

PR #170, opened 2024-02-21
Use margin in joining catalogs

smcguire-cmu:
Joining catalogs together now uses the margin in the right catalog, in the same way that cross matching does.

In testing, I found a bug in the way we were joining the right and right margin catalog, that for the margin pixels the `Norder`, `Npix`, and `Dir` columns use the columns they are in for the margin cache and not that match the ra and dec columns. So this now changes that behavior to use the values that match their original pixel.


tests/conftest.py
```diff
@@ -116,6 +140,26 @@ def small_sky_order1_catalog(small_sky_order1_dir):
     return lsdb.read_hipscat(small_sky_order1_dir)
 
 
+@pytest.fixture
+def small_sky_order1_source_catalog(small_sky_order1_source_dir, small_sky_order1_source_margin_catalog):
```
delucchi-cmu, line 144:
```suggestion
def small_sky_order1_source_with_margin(small_sky_order1_source_dir, small_sky_order1_source_margin_catalog):
```

I know this fixture already has a long name, but it would make the tests a lot clearer that this catalog already has a margin included.

tests/data/small_sky_to_o1source_soft/catalog_info.json
delucchi-cmu, line 1: Can you add a little README here? In particular, by "soft", do you just mean that it's an association table that doesn't contain leaf files?

src/lsdb/dask/merge_catalog_functions.py
```diff
@@ -18,6 +19,37 @@
...
+    Returns:
+        The concatenated dataframe with the partition on top and the margin on the bottom
+    """
+    hive_columns = [
+        PartitionInfo.METADATA_ORDER_COLUMN_NAME,
+        PartitionInfo.METADATA_DIR_COLUMN_NAME,
+        PartitionInfo.METADATA_PIXEL_COLUMN_NAME,
+    ]
+    margin_filtered = None
+    if margin is not None:
```
delucchi-cmu, line 40:
```suggestion
    if margin is not None:
      # Remove the Norder/Dir/Npix columns (used only for partitioning the margin itself), 
      # and rename the margin_Norder/Dir/Npix to take their place.
```

This is what I thought was happening in this block.

src/lsdb/dask/merge_catalog_functions.py
```diff
@@ -18,6 +19,37 @@
...
+        margin_columns_no_hive = [col for col in margin.columns if col not in hive_columns]
+        rename_columns = {
+            f"margin_{PartitionInfo.METADATA_ORDER_COLUMN_NAME}": PartitionInfo.METADATA_ORDER_COLUMN_NAME,
+            f"margin_{PartitionInfo.METADATA_DIR_COLUMN_NAME}": PartitionInfo.METADATA_DIR_COLUMN_NAME,
+            f"margin_{PartitionInfo.METADATA_PIXEL_COLUMN_NAME}": PartitionInfo.METADATA_PIXEL_COLUMN_NAME,
+        }
+        margin_renamed = margin[margin_columns_no_hive].rename(columns=rename_columns)
+        margin_filtered = margin_renamed[right_columns]
+    joined_df = pd.concat([partition, margin_filtered]) if margin_filtered is not None else partition
```
delucchi-cmu, line 49: This method would be easier to read with return early, e.g.:

```
def do_stuff():
  if margin is None:
    return partition
  hive_columns = ...
  ...
  return pd.concat([partition, margin_filtered)]
```

src/lsdb/dask/join_catalog_data.py
```diff
@@ -210,16 +243,24 @@ def join_catalog_data_through(
             association.hc_structure.catalog_info.join_column,
             suffixes,
         )
-    alignment = align_trees(
-        left.hc_structure.pixel_tree, right.hc_structure.pixel_tree, alignment_type=PixelAlignmentType.INNER
-    )
+
+    if right.margin is None:
+        warnings.warn("Right catalog does not have a margin cache. Results may be inaccurate", RuntimeWarning)
```
delucchi-cmu, line 248: These warnings are being thrown a lot in the unit tests. Are you planning to add margins to additional data sets (maybe as a follow-up), or expect the warnings in the unit tests?

tests/data/small_sky_to_o1source_soft/catalog_info.json
smcguire-cmu: Yep, that's what I meant

src/lsdb/dask/merge_catalog_functions.py
smcguire-cmu: Ah that's nicer!

delucchi-cmu (approved):

smcguire-cmu:
Err, something messed up trying to resolve conflicts there

PR #175, opened 2024-02-23
Generate margin catalog on `from_dataframe`

### LINKED ISSUE #149: Generate margin catalog with `from_dataframe`
Currently, if `from_dataframe` is called, the catalog will be created without a margin. To use it in a crossmatch with a margin, the user would have to save the catalog and run hipscat_import to generate the margin. There should be an optional argument to also generate the margin cache at a given radius, and return the catalog with its margin cache.

camposandro:
Adds margin generation when loading a catalog with `from_dataframe`. Closes #149.


src/lsdb/loaders/dataframe/from_dataframe_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    """Append partitioning information to a HEALPix dataframe
+
+    Args:
+        dataframe (pd.Dataframe): A HEALPix's pandas dataframe
+        pixel (HealpixPixel): The HEALPix pixel for the current partition
+
+    Returns:
+        The dataframe for a HEALPix, with data points and respective partition information.
+    """
+    dataframe["Norder"] = pixel.order
```
delucchi-cmu, line 43: You're mixing use of `"Norder"` and `PartitionInfo.METADATA_ORDER_COLUMN_NAME` quite a bit in this PR. I think we should pick a convention and stick with it.

delucchi-cmu (approved):

PR #183, opened 2024-02-27
Create validation method.

### LINKED ISSUE #161: Raise an exception for missed / mismatched margin caches
Currently, `.crossmatch` warns when right catalog doesn't have margin cache. I believe that we should:
1. raise an exception instead (maybe with an optional keyword argument converting it back to a warning);
2. check if search radius is not larger than the margin.

delucchi-cmu:
Closes #161 .

Creates a new validation method that cross match algorithms must implement. 

This will validate the structure of the dataframes (using the metadata dataframe), and any arguments that will be passed to the crossmatching method.


src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
```diff
@@ -51,12 +54,22 @@ def __init__(
...
         self.right_metadata = right_metadata
+        self.right_margin_hc_structure = right_margin_hc_structure
         self.suffixes = suffixes
 
     @abstractmethod
     def crossmatch(self) -> pd.DataFrame:
         """Perform a crossmatch"""
 
+    @abstractmethod
+    def validate(self):
```
hombit, line 65: Should they both be `(self, *args, **kwargs)` or something like that?

src/lsdb/core/crossmatch/kdtree_match.py
```diff
@@ -15,10 +16,42 @@ class KdTreeCrossmatch(AbstractCrossmatchAlgorithm):
...
+        if self.left_metadata.catalog_info.ra_column not in column_names:
+            raise ValueError(f"left table must have column {self.left_metadata.catalog_info.ra_column}")
+        if self.left_metadata.catalog_info.dec_column not in column_names:
+            raise ValueError(f"left table must have column {self.left_metadata.catalog_info.dec_column}")
+
+        column_names = self.right.columns
+        if self.right_metadata.catalog_info.ra_column not in column_names:
+            raise ValueError(f"right table must have column {self.right_metadata.catalog_info.ra_column}")
+        if self.right_metadata.catalog_info.dec_column not in column_names:
+            raise ValueError(f"right table must have column {self.right_metadata.catalog_info.dec_column}")
```
smcguire-cmu, line 39: Could we make the method in the abstract class not abstract and use this as a default so a custom algorithm doesn't need to implement a validate?

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
delucchi-cmu: Added kwargs in a few places, until pylint/mypy stopped complaining.

smcguire-cmu (approved):
Looks good thanks!

PR #188, opened 2024-02-28
Numpy tricks for box_search.py

hombit:
1. Default `mask` array had wrong dtype - it would cause a dataframe where all the elements are the second row of the original data frame
2. Use `np.asarray` to prevent data copy


camposandro (approved):
Thank you, @hombit!

PR #205, opened 2024-03-05
Prune empty partitions from catalog

### LINKED ISSUE #200: Prune empty partitions with explicit compute
Enable pruning empty partitions by explicitly checking if the partitions are empty. The only way to do it is to calculate the partition lengths (load them into memory) to check which ones are empty.

camposandro:
Prune empty partitions from catalogs by explicitly computing the partition lengths. Closes #200.


smcguire-cmu (commented):
Looks good! Just one thing I think would be great to add

smcguire-cmu (approved):

PR #207, opened 2024-03-05
Change _DIST to _dist_arcsec

### LINKED ISSUE #141: Change _DIST units to arcseconds
After we changed cross-match radius units to arcseconds, it has a good sense to also change `_DIST` column units from degrees to arcseconds
https://github.com/astronomy-commons/lsdb/blob/1ca91c30312a32290b43e8d1b2fecb41af02db1a/src/lsdb/core/crossmatch/kdtree_match.py#L64

smcguire-cmu:
Closes #141 by changing the "_DIST" column to "_dist_arcsec" in the KDTree crossmatch result as discussed.


camposandro (approved):
Everything looks right!

PR #208, opened 2024-03-06
Save margin catalog to disk

### LINKED ISSUE #206: Save margin catalog with `to_hipscat`
**Feature request**

There should be an option to save a margin catalog to disk. There is already a `to_hipscat` method for the **lsdb.Catalog**, so we might just need to refactor it and include it in the **HealpixDataset** superclass.

https://github.com/astronomy-commons/lsdb/blob/d7d3ac8fbcc0131f96c878567c56bf080a790d38/src/lsdb/catalog/catalog.py#L327-L344

**Before submitting**
Please check the following:

- [X] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

camposandro:
Refactors the `to_hipscat` method to enable saving a margin catalog to disk. Closes #206.

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -1,14 +1,13 @@
 from __future__ import annotations
 
 import dataclasses
-from typing import Any, Dict, List, Tuple, Type, Union
+from typing import List, Tuple, Type
 
 import dask.dataframe as dd
 import hipscat as hc
 import pandas as pd
 from hipscat.pixel_math.polygon_filter import SphericalCoordinates
 
-from lsdb import io
 from lsdb.catalog.association_catalog import AssociationCatalog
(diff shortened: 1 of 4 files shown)


delucchi-cmu (approved):
To confirm, the expected usage would be something like:

```
my_data = ...
catalog = lsdb.from_dataframe()
catalog.to_hipscat("/path/to/new/catalog", catalog_name ="base_catalog")
catalog.margin.to_hipscat("path/to/new/catalog", catalog_name="margin_catalog")
```

I think I can see why the way you've implemented is better, but I had originally thought the API would look something like:

```
my_data = ...
catalog = lsdb.from_dataframe()
catalog.to_hipscat("/path/to/new/catalog", catalog_name ="base_catalog", margin_catalog_name="margin_catalog")
```

PR #209, opened 2024-03-06
Add option for coarse spatial filtering

### LINKED ISSUE #201: Add ability to do coarse spatial filtering
In the case where one wants do the spatial filtering and do it quickly - allow an user to skip filtering down to exact coordinates provided and keep all the pixels that have any area within the specified spatial region.

camposandro:
Provides coarse filtering option for spatial searches. Coarse filtering allows to keep the partitions for the region provided by the user (which is computed lazily) without filtering the data points they contain to the exact region coordinates. The coarse option speeds up searches for all those science cases that only require a rough region filtering. Closes #201.


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -114,20 +114,25 @@ def query(self, expr: str) -> Self:
...
         Returns:
             A tuple containing a dictionary mapping pixel to partition index and a dask dataframe
             containing the search results
         """
         partitions = self._ddf.to_delayed()
         targeted_partitions = [partitions[self._ddf_pixel_map[pixel]] for pixel in filtered_pixels]
-        filtered_partitions = [search.search_points(partition) for partition in targeted_partitions]
+        filtered_partitions = [
+            search.search_points(partition) if fine else partition for partition in targeted_partitions
+        ]
```
smcguire-cmu, line 135:
```suggestion
        filtered_partitions = [
            search.search_points(partition) for partition in targeted_partitions
        ] if fine else targeted_partitions
```

smcguire-cmu: Prevents unnecessary list comprehension if not fine

smcguire-cmu (commented):
Looks fine code wise, but I'm not sure about having fine False by default. I'm guessing you've spoken to the project scientists about this, and I get that for our use cases that we mostly just want coarse filtering, but I worry that most people will not read the documentation and assume they're doing a fine filter. The healpix fine/coarse thing isn't particularly intuitive at first glance and setting fine to False is just a slight speedup in compute time.

camposandro:
@smcguire-cmu I agree, I don't find it too intuitive to have the coarse filtering by default. The last time we discussed this I was left with the impression that coarse was the way to go because what most science cases look for is a rough sky region filtering, but I am comfortable doing otherwise. @nevencaplar @hombit do you have any strong preference?

camposandro:
@smcguire-cmu the default filtering strategy has been changed to fine.

smcguire-cmu (approved):

PR #210, opened 2024-03-06
Notebook for generating test data

delucchi-cmu:
## Change Description

This is a deceptively large PR, but not all that much is changing!!

With recent changes to the format and to the importer, this re-generates ALL of the unit test catalog data. This does this through the new jupyter notebook. After this change, if you run the entire notebook, the only changed files should be `provenance_info.json` file changes (which are noisy changes). Some catalogs had their columns renamed manually on the previous imports, so I've just changed the expected columns in test code.

Also sneaks in some other test-data-related quality of life things:
- ignores `point_map.fits` files from source control. these are large and are not tested in this repo.
- moves CSV files with expected crossmatch results into their own directory


smcguire-cmu (approved):
Looks good! I like that SOAP now uses the right catalog_info parameters, last time I had to add those in manually but this is so much better thanks!

PR #216, opened 2024-03-08
Patch margin cache generation

camposandro:
This pull request patches some inconsistencies detected in the `from_dataframe` margin cache generation.

These changes focus on:
- Generating the margin shards for each HEALPix partition, excluding those that are empty;
- Sorting the catalog pixel map by `_hipscat_index`;
- Making sure the margin catalog is set to None in case there are no points in the margin at all. This way we make sure we don't use it in other operations afterward (e.g. in catalog `query`).

I also included some additional test data (*small_sky_order3_source* catalog and *small_sky_order3_source_margin*) using a different margin order and threshold for robustness.


tests/data/generate_data.ipynb
delucchi-cmu, line 1: =D

delucchi-cmu (approved):

maxwest-uw (approved):

PR #230, opened 2024-03-12
Change metadata loading method to use right file

smcguire-cmu:
uses `_common_metadata` instead of `_metadata`, which gives a large performance improvement to loading catalogs


delucchi-cmu (approved):
:1st_place_medal:

PR #231, opened 2024-03-12
Load catalog subset at the I/O level

### LINKED ISSUE #204: Loading Subsets at the I/O level
** Feature request**
This is related to #90.

Currently, for generating subsets of data we abide by a workflow structure that looks like:
1. "Load" all data
2. Filter using some mechanism (e.g. box search, cone search)

This is reasonable for most use cases, as Dask will only actively read data from the partitions used in the filter (at least in the case of box search). However, on the flip side this filter does not prevent Dask from constructing an initial layer of the task graph that considers all partitions in the dataset. This imposes an upper-bound on the scale that LSDB can work with very large datasets, as at a certain number of partitions the task graph will be so large that prevents even a subset from being loaded in efficiently.

It would be nice to have some tooling available to use the hipscat information to filter down to a small subset of the data before or as part of the `read_hipscat` call, so as to prevent Dask from needing to construct the full first layer of read_parquet calls.

I'm not entirely sure if this is feasible, as in the case of many partitions this initial step will need to take on the burden of filtering on it's own and maybe this ends up having the same effect as kicking it to Dask. Maybe something even to the level of grabbing the first n files would be helpful for users who want to do workflow validation. Really rough API example:
```
import lsdb

small_fileset = lsdb.create_hipscat_subset(hipscat_path, order=5, nfiles=20)
#or small_fileset = lsdb.create_hipscat_subset(hipscat_path, method=cone_search, ra=40, dec=11, radius=1)
lsdb.read_hipscat(small_fileset)
```

**Before submitting**
Please check the following:

camposandro:
Implements `read_hipscat_subset`, a wrapper to the `read_hipscat` method that allows to select partitions ahead of time according to a spatial filter or a pre-defined number of pixels. The goal is to prevent Dask from trying to figure out which parquet files should be read or not for our workflow, reducing the size of the task graph. This should speed-up loading time, in particular for catalogs with a very large number of partitions.

**Note:** Building the catalog / filter pixel trees and reading the parquet metadata are still the most time-consuming operations. The optimizations implemented by Sean around the pixel tree generation in conjunction with this selective partition loading should help to improve the loading time significantly. Some benchmarking is still being performed.

Closes #204.

### DIFF
--- a/src/lsdb/loaders/hipscat/read_hipscat.py
+++ b/src/lsdb/loaders/hipscat/read_hipscat.py
@@ -124,6 +124,7 @@ def read_hipscat_subset(
     pixels_to_load = None
     hc_structure = HCHealpixDataset.read_from_hipscat(path, storage_options)
     pixels = hc_structure.get_healpix_pixels()
+
     if search_filter is not None and n_pixels is not None:
         raise ValueError("Use only one of `search_filter` or `n_pixels`")
     if search_filter is not None:
@@ -131,6 +132,10 @@ def read_hipscat_subset(
     elif n_pixels is not None:
         order = order if order is not None else hc_structure.partition_info.get_highest_order()
         pixels_to_load = _find_first_pixels_of_order(pixels, order, n_pixels)
+
+    if pixels_to_load is not None and len(pixels_to_load) == 0:
+        raise ValueError("There are no partitions for the specified subset")
+
     return read_hipscat(
         path=path,
         catalog_type=catalog_type,
@@ -155,6 +160,6 @@ def _get_dataset_class_from_catalog_info(
 def _find_first_pixels_of_order(pixels: List[HealpixPixel], order: int, n_pixels: int) -> List[HealpixPixel]:
     """Retrieves the first `n_pixels` of the catalog at the specified `order`"""
     pixels_of_order = [pixel for pixel in pixels if pixel.order == order]
-    if n_pixels > len(pixels_of_order):
-        warnings.warn(f"There are less than {n_pixels} at order {order}. Loading all pixels.", RuntimeWarning)
+    if 0 < len(pixels_of_order) < n_pixels:
+        warnings.warn(f"There are less than {n_pixels} pixels at order {order}. Loading all.", RuntimeWarning)
     return pixels_of_order[:n_pixels]
--- a/src/lsdb/loaders/hipscat/read_hipscat.py
+++ b/src/lsdb/loaders/hipscat/read_hipscat.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
+import warnings
 from typing import Any, Dict, List, Type, Union
 
 import hipscat as hc
@@ -76,11 +77,13 @@ def read_hipscat_subset(
     catalog_type: Type[Dataset] | None = None,
     storage_options: dict | None = None,
     search_filter: AbstractSearch | None = None,
-    n_files: int | None = None,
     order: int | None = None,
+    n_pixels: int | None = None,
     **kwargs,
-):
+) -> Dataset:
     """Load a catalog subset from a HiPSCat formatted catalog.
+
+    To be used with a filter or (order, n_files), but not both.
 
     Typical usage example, where we load a catalog from a cone search:
         lsdb.read_hipscat_subset(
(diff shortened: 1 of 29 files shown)


src/lsdb/loaders/hipscat/read_hipscat.py
```diff
@@ -66,6 +73,79 @@ def read_hipscat(
...
 
+def read_hipscat_subset(
+    path: str,
+    catalog_type: Type[Dataset] | None = None,
+    storage_options: dict | None = None,
+    search_filter: AbstractSearch | None = None,
+    order: int | None = None,
+    n_pixels: int | None = None,
+    **kwargs,
+) -> Dataset:
```
smcguire-cmu, line 84: I'm not sure we need a separate read method for the subset. Could we add the search_filter option to `read_hipscat` instead?

src/lsdb/loaders/hipscat/read_hipscat.py
```diff
@@ -66,6 +73,79 @@ def read_hipscat(
...
+            cannot allow a return type specified by a loaded value, so to use the correct return
+            type for type checking, the type of the catalog can be specified here. Use by specifying
+            the lsdb class for that catalog.
+        storage_options (dict): Dictionary that contains abstract filesystem credentials
+        search_filter (Type[AbstractSearch]): The filter method to be applied to the catalog.
+            By default, all files are considered.
+        n_pixels (int): The number of files to read, in case no filter was specified.
+            By default, all files are considered.
+        order (int): The order of the files to read, in case n_files was specified.
+            By default, the largest catalog order is used.
```
smcguire-cmu, line 119: Using order like this feels unintuitive to me. I'd expect that by default setting `n_pixels` would get the first n pixels, not the first n pixels of the highest order.

src/lsdb/loaders/hipscat/read_hipscat.py
```diff
@@ -66,6 +73,79 @@ def read_hipscat(
...
+            filter=lsdb.core.search.ConeSearch(ra, dec, radius_arcsec),
+        )
+
+    Typical usage example, where we load a catalog from a pre-defined number of files:
+         lsdb.read_hipscat_subset(
+            path="./my_catalog_dir",
+            catalog_type=lsdb.Catalog,
+            columns=["ra","dec"],
+            order=5,
+            n_pixels=10,
```
delucchi-cmu, line 103: This search syntax is not intuitive to me. I think it's because `Npix` is a very specific thing in HiPS. it's also bizarre to me to get those partitions returned that are at a specific order.

alternative a. rename to `num_partitions`?
alternative b. provide some kind of `partition_list`?
alternative c. remove this style of filtering for this PR so we can create a separate discussion about it, because i have even more alternatives and i don't want to de-rail the foundational work.

src/lsdb/loaders/hipscat/read_hipscat.py
```diff
@@ -66,6 +73,79 @@ def read_hipscat(
...
+        order (int): The order of the files to read, in case n_files was specified.
+            By default, the largest catalog order is used.
+        **kwargs: Arguments to pass to the read hipscat call
+
+    Returns:
+        Catalog object loaded from the given parameters
+    """
+    pixels_to_load = None
+    hc_structure = HCHealpixDataset.read_from_hipscat(path, storage_options)
+    pixels = hc_structure.get_healpix_pixels()
```
smcguire-cmu, line 127: I think the best way of structuring this to avoid loading the hipscat catalog twice would be to move the filtering logic into the `AbstractCatalogLoader` so that you can perform the filtering on the hc_structure after it's loaded there and use that as the hc_structure instead of making it here then making it again there.

src/lsdb/loaders/hipscat/read_hipscat.py
camposandro: I went with alternative c. to make this PR focus solely on the spatial filtering!

src/lsdb/loaders/hipscat/read_hipscat.py
camposandro: Thanks, good catch. This call is expensive and performing it twice was hindering the performance significantly!

delucchi-cmu (approved):

PR #235, opened 2024-03-14
Use custom representation for catalog.

### LINKED ISSUE #126: Catalog description when viewing it could be improved to avoid confusion
Catalog print out says ``Dask DataFrame Structure''
Screenshot 2024-01-22 at 3 14 23 PM

but at the same time it is not really Dask DataFrame, which leads to confusion with users. @wilsonbb mentioned to me that the text at the top can be changed - I suggest to make it more suggestive, i.e., lsdb Catalog.

delucchi-cmu:
Closes #126 

- includes the name of the catalog
- doesn't issue a warning from within dask

image


smcguire-cmu (approved):

smcguire-cmu:
This looks nice!

PR #238, opened 2024-03-15
Filter based on HEALpix order

### LINKED ISSUE #237: Add functionality to select only pixels above a certain order.
Select only pixels above a certain order (roughly corresponding to the above certain density on the sky). This is a functionality that can be, e.g., used to select objects within the galactic plane.

ykwang1:
Add method to filter pixels by minimum and maximum order to extract areas of higher or lower density, such as the galactic plane.

Closes #237


delucchi-cmu (approved):
Looks great!

PR #243, opened 2024-03-19
Update unit test data

### LINKED ISSUE #242: Slow unit tests
**Bug report**

These are some slow running unit tests that should be improved. I think a lot of this comes from doing pandas `iter_rows` to make assertions about the data that is now prohibitively slow after small_sky_order1_source has a lot more data:

didn't finish after 5 mins     `tests/lsdb/catalog/test_join.py::test_join_association_source_margin`
22.67s call     `tests/lsdb/catalog/test_cone_search.py::test_cone_search_filters_correct_points_margin`
13.88s call     `tests/lsdb/catalog/test_join.py::test_small_sky_join_small_sky_order1_source`
7.79s call     `tests/lsdb/loaders/dataframe/test_from_dataframe.py::test_from_dataframe_small_sky_order3_source_with_margins`
2.61s call     `tests/lsdb/loaders/dataframe/test_from_dataframe.py::test_from_dataframe_small_sky_source_with_margins`
1.63s call     `tests/lsdb/catalog/test_join.py::test_join_association`

smcguire-cmu:
Regenerates the unit test data but with `small_sky_source_order1` now being partitioned by the source ra and dec instead of being copartitioned with the objects using the objects ra and dec. Its margin and association catalogs are also updated.

Changing small_sky_source_order1 to include many more points introduced slow downs in some unit tests. This PR also fixes these, by replacing pandas iteration with np checks where possible, storing an expected result to check against for a cone search test, and removing the `test_from_dataframe_small_sky_order3_source_with_margins` unit test. (The computation is a little large for a unit test now.) Fixes #242


tests/conftest.py
```diff
@@ -233,6 +233,21 @@ def xmatch_mock(xmatch_expected_dir):
...
 
 
+@pytest.fixture
+def cone_search_expected_dir(test_data_dir):
+    return os.path.join(test_data_dir, "raw", "cone_search_expected")
+
+
+@pytest.fixture
+def cone_search_expected(cone_search_expected_dir):
+    return pd.read_parquet(os.path.join(cone_search_expected_dir, "catalog.parquet"))
```
delucchi-cmu, line 243: Could we make these CSV files instead, just for ease of human readability?  Or just a text file with the `source_id` field.

delucchi-cmu (approved):

PR #245, opened 2024-03-20
add skymap_histogram function and  skymap at specified order

smcguire-cmu:
Adds the `skymap_histogram` function that returns a numpy array of the histogram instead of plotting. Also adds an `order` parameter to the skymap functions that allow for the skymap to be computed at a fixed order, and a `default_value` parameter that allows the user to specify the value used for empty partitions in the histogram.

Changes the skymap function to use the same `projection` parameter to choose a projection function that the `hipscat` plotting libraries use.


src/lsdb/core/plotting/skymap.py
```diff
@@ -1,32 +1,61 @@
...
 
-def plot_skymap(pixel_map: Dict[HealpixPixel, Any], **kwargs):
-    """Plot a map of healpix_pixels with values on a mollweide projection.
+def compute_skymap(pixel_map: Dict[HealpixPixel, Any], order: int | None = None, default_value: Any = 0.0):
+    """Returns a histogram map of healpix_pixels to values.
 
     Args:
         pixel_map(Dict[HealpixPixel, Any]): A dictionary of healpix pixels and their values
-        kwargs: Additional keyword arguments to pass to `healpy.mollview`
+        order (int): The order to make the histogram at
```
camposandro, line 39: It would be useful to have the default values for `order` and `default_value` in the docstring

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -205,41 +206,104 @@ def _get_non_empty_partitions(self) -> Tuple[List[HealpixPixel], List[Delayed]]:
...
-            pixel: delayed(func)(partitions[index], pixel, **kwargs)
-            for pixel, index in self._ddf_pixel_map.items()
-        }
+        if order is None:
+            results = {
+                pixel: delayed(func)(partitions[index], pixel, **kwargs)
+                for pixel, index in self._ddf_pixel_map.items()
+            }
+        else:
+            if order < self.hc_structure.pixel_tree.get_max_depth():
```
camposandro, line 232: I do not understand if it's useful to allow for fixed orders higher than the catalog's maximum order. In practice, won't all those plots look the same as if they were computed at the maximum order?

camposandro (commented):
The code looks great!!

src/lsdb/core/plotting/skymap.py
```diff
@@ -1,32 +1,61 @@
...
+    dorders = order - hp_orders
+    values = [pixel_map[x] for x in pixels]
+    starts = hp_pixels << (2 * dorders)
+    ends = (hp_pixels + 1) << (2 * dorders)
 
     def set_values(start, end, value):
         img[np.arange(start, end)] = value
 
-    np.vectorize(set_values)(starts, ends, values)
+    for s, e, v in zip(starts, ends, values):
```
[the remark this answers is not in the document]
smcguire-cmu: I've seen mixed things on this, but I think the answer is that for ufuncs like set_values it basically just wraps the for loop anyway. I haven't profiled to see if there's a performance impact. The reason this is a for loop is that np.vectorize doesn't work for when values is a list of arrays of different dimensions, which it will be when a fixed order is set.

camposandro (approved):

Issue #151, opened 2024-02-09
More options for nearest neighbor selection for KdTreeCrossmatch


hombit:
Currently, we have two ways to control neighbor selection with the KdTreeCrossmatch algorithm: the maximum number of neighbors to select `n_neighbors` and the maximum radius `radius_arcsec`. However, it would be beneficial to add more flexibility:

- [ ] Set a minimum radius and/or skip the first nth matches. This would be very useful for catalog self-matching tasks.
- [ ] Support a "left" join in addition to the "inner" join we have now, so that rows from the left table that do not match are kept in the result.

I propose that we keep this issue open for discussions while creating separate issues for individual tasks.

nevencaplar:
@hombit Can you just explain a little bit what you mean by ``set a minimum radius'' - skip a match if it it below a certain distance? Is this so that you ensure that you are matching to a different object in a crossmatch, making sure to avoid the same object in the second catalog?

hombit:
@nevencaplar, yes, I meant this. I was thinking about catalog self-matching. For example, you'd like to find wide binaries and looking for pairs of stars in 2''. However, you wouldn't like to have objects to match to themselves, so you would set `nthmatch=2`. However, it wouldn't work with some catalogs having the same sky source to be represented by few rows, for instance "source" catalogs, or catalogs with duplicated objects such as PS1 DR2.

nevencaplar:
What is happening here? I see that this ticket has a pull from April? @camposandro ?

camposandro:
We left #271 on hold during the feature freeze leading to the IVOA demos and I didn't have the chance to look back into it. I need to resolve the merge conflicts and request a new review.

PR #246, opened 2024-03-20
Add minimum radius option to KDTree crossmatch

camposandro:
Addresses the first feature request in ticket #151, allowing to set a minimum radius for the KDTree crossmatch.


src/lsdb/core/crossmatch/kdtree_match.py
```diff
@@ -17,25 +19,36 @@ class KdTreeCrossmatch(AbstractCrossmatchAlgorithm):
...
                 raise ValueError("Cross match radius is greater than margin threshold")
+            if self.right_margin_hc_structure.catalog_info.margin_threshold < min_radius_arcsec:
+                raise ValueError("Cross match minimum radius is greater than margin threshold")
 
     # pylint: disable=unused-argument
     def crossmatch(
         self,
         n_neighbors: int = 1,
         radius_arcsec: float = 1,
+        min_radius_arcsec: float = 0,
```
smcguire-cmu, line 51: Can we have a unit test for a self match to make sure we're not ignoring matches with 0 distance between themself?

src/lsdb/core/crossmatch/kdtree_match.py
```diff
@@ -139,3 +153,65 @@ def _lon_lat_to_xyz(lon: npt.NDArray[np.float64], lat: npt.NDArray[np.float64])
...
+    max_distance: float,
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Finds `n_neighbors` within a distance range for all points in a pair of partitions"""
+    left_tree = KDTree(left_xyz, compact_nodes=True, balanced_tree=True, copy_data=False)
+
+    # Find the number of neighbors within the minimum distance threshold
+    len_too_close_neighbors = np.zeros(left_xyz.shape[0], dtype=np.int64)
+    left_indices_too_close = left_tree.sparse_distance_matrix(
+        tree, max_distance=min_distance, output_type="ndarray"
+    )["i"]
```
smcguire-cmu, line 187: comment to explain what `["I"]` does here pls

src/lsdb/core/crossmatch/kdtree_match.py
```diff
@@ -139,3 +153,65 @@ def _lon_lat_to_xyz(lon: npt.NDArray[np.float64], lat: npt.NDArray[np.float64])
...
+    unique, counts = np.unique(left_indices_too_close, return_counts=True)
+    len_too_close_neighbors[unique] = counts
+
+    # Make sure we don't ask for more neighbors than there are points
+    n_neighbors_to_request = min(n_neighbors + max(len_too_close_neighbors), len(right_xyz))
```
smcguire-cmu, line 192: I think the mask logic below fails if there are fewer points than we want to ask for

hombit (commented):
I believe this change is large enough to move it to a separate cross-matching algorithm

src/lsdb/core/crossmatch/kdtree_match.py
```diff
@@ -17,25 +19,36 @@ class KdTreeCrossmatch(AbstractCrossmatchAlgorithm):
...
+        self,
+        n_neighbors: int = 1,
+        radius_arcsec: float = 1,
+        min_radius_arcsec: float = 0,
+        require_right_margin=True,
+        **kwargs,
+    ):
         super().validate()
         validate_radius(radius_arcsec)
+        if min_radius_arcsec < 0:
```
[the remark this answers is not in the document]
camposandro: `validate_radius` ensures that radii are positive but here the minimum radius might default to 0.

src/lsdb/core/crossmatch/kdtree_match.py
camposandro: That's right! I pushed a fix and added a unit test to make sure the mask is successfully generated for this case.

src/lsdb/core/crossmatch/kdtree_match.py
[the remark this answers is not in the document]
smcguire-cmu: ok but why less than 0?

tests/data/raw/xmatch_expected/xmatch_correct_002_005.csv
```diff
@@ -0,0 +1,31 @@
+,ss_id,xmatch_id,dist
```
smcguire-cmu, line 1: How are these files generated?

camposandro: They were generated using a brute-force approach that computes the distances between the pairs of points formed between `small_sky` and `small_sky_xmatch`. 

Of those points, those that are within the `min_radius_arcsec` and `radius_arcsec` are kept and sorted by ascending order of separation. Finally, we select up to `n_neighbors`.

src/lsdb/core/crossmatch/kdtree_match.py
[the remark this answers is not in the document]
camposandro: It's just a complementary check to warn users that this radius must be non-negative. I improved the error message.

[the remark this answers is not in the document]
smcguire-cmu: I still think this is a bit difficult to read, calling a separate validation function after already performing the validation. I think raising the error instead of calling validate radius is better here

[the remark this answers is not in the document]
camposandro: Yes, I agree! That's what I did really, for some reason this view of the code is outdated.

smcguire-cmu (approved):

PR #274, opened 2024-04-11
Optimize skymap at high fixed order

### LINKED ISSUE #266: Simple functions take huge amount of time on few partitions
**Bug report**

We tried running the following, relatively simple code 
code_example

where Trilegal is a catalog with around 4500 partitions, we run on 96 cores with plenty of memory. The code goes quickly through most of the partitions, but then it has on last handful; these few partitions take more than the bulk of the partitions!  See the figures below showing how long it takes to get through last few of them!

dashboard
dashboard_process

I am also adding a figure that shows that partitions are roughly of the same size, i.e., it doesn't take too long for few of them because they are too big. 
sizes

It is possible that this is a problem with this particular catalog, but I believe that this might be a more general problem. It might not be LSDB problem, but an issue from when we are creating catalog as well (this one was done with hipscat-import).

1. Determine if this problem is reproducible with other catalogs (e.g., GAIA, ZTF)
2. If yes, identify and fix the issue.

**Before submitting**
Please check the following:

- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.

smcguire-cmu:
Changes vectorized filtering to a groupby. Fixes #266

### DIFF
--- a/src/lsdb/core/plotting/skymap.py
+++ b/src/lsdb/core/plotting/skymap.py
@@ -8,8 +8,6 @@ import pandas as pd
 from dask import delayed
 from hipscat.pixel_math import HealpixPixel, hipscat_id_to_healpix
 
-from lsdb.dask.merge_catalog_functions import filter_by_hipscat_index_to_pixel
-
 
 @delayed
 def perform_inner_skymap(
--- a/src/lsdb/core/plotting/skymap.py
+++ b/src/lsdb/core/plotting/skymap.py
@@ -15,6 +15,7 @@ def perform_inner_skymap(
     func: Callable[[pd.DataFrame, HealpixPixel], Any],
     pixel: HealpixPixel,
     target_order: int,
+    default_value: Any = 0,
     **kwargs,
 ) -> np.ndarray:
     """Splits a partition into pixels at a target order and performs a given function on the new pixels"""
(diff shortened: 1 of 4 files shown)


src/lsdb/core/plotting/skymap.py
```diff
@@ -43,6 +45,9 @@ def compute_skymap(
     """
 
     pixels = list(pixel_map.keys())
+    if len(pixels) == 0:
```
camposandro, line 48: Looks good! Can you just add a unit test for this case?

tests/lsdb/catalog/test_catalog.py
```diff
@@ -345,10 +343,16 @@ def func(df, healpix):
...
+def test_skymap_histogram_empty(small_sky_order1_catalog):
+    def func(df, healpix):
+        return len(df) / hp.nside2pixarea(hp.order2nside(healpix.order), degrees=True)
+
+    expected_img = np.full(12, 1)
+    img = small_sky_order1_catalog.cone_search(0, 0, 1).skymap_histogram(func, default_value=1)
+    assert (img == expected_img).all()
+
 
+def test_skymap_histogram_order_default(small_sky_order1_catalog):
```
camposandro, line 355: It looks like you're missing the default value when initializing the `pixel_map` (line 362).

smcguire-cmu: Thanks, I changed the test to fix this

camposandro (approved):

PR #299, opened 2024-04-26
Don't re-execute slow notebooks.

### LINKED ISSUE #259: Docs notebook rendering for longer notebooks
Currently we can't have longer running tasks in notebooks that are part of the documentation, since they are run in the Github actions and RTD runners to generate the docs pages, and these will timeout for large notebooks, and mean that the actions take very long in each PR.

It would be nice if we could have some way to use pre-rendered notebooks in the docs in certain cases, but this would involve modifying how the docs notebook building works and changing the empty notebook output precommit check.

delucchi-cmu:
Create directory for slower-running, pre-executed notebooks that are not cleared in the pre-commit stage and adds metadata to the notebook just to be sure it won't be run in CI:

```
   "nbsphinx": {
    "execute": "never"
   },
```

Closes #259


drewoldag (approved):
Looks good. I'm excited about including this in the PPT.

PR #305, opened 2024-05-02
Replace `.values` accessor with `.to_numpy()`

### LINKED ISSUE #286: Change uses of pandas `.values`
**Bug report**

We use pandas `df.values` in many places across the codebase, in LSDB and HiPSCat. This should be changed, as the `values` method [has a warning in the pandas documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.values.html), as it can return different types depending on the backend used.

We should also have a bigger conversation around supported backends. We don't test across different supported backends, only the default numpy, and in some places we assume this is the backend like with using `values`.

camposandro:
The DataFrame `.values` accessor returns different data types depending on the backend used to load the data. Going forward we will be supporting the specification of different backends and, therefore, we will be accessing the values in their numpy representation. Closes #286.


src/lsdb/core/plotting/skymap.py
```diff
@@ -19,19 +19,19 @@ def perform_inner_skymap(
...
-        p = hipscat_id_to_healpix([df.index.values[0]], target_order=target_order)[0]
+        p = hipscat_id_to_healpix([df.index.to_numpy()[0]], target_order=target_order)[0]
         return func(df, HealpixPixel(target_order, p), **kwargs)
 
     gb = partition.groupby(order_pixels, sort=False).apply(apply_func)
     delta_order = target_order - pixel.order
     img = np.full(1 << 2 * delta_order, fill_value=default_value)
     min_pixel_value = pixel.pixel << 2 * delta_order
-    img[gb.index.values - min_pixel_value] = gb.values
+    img[gb.index.to_numpy() - min_pixel_value] = gb.to_numpy()
```
smcguire-cmu, line 34: This could probably do with checking for null values here, that `gb.to_numpy()` is a user column that could have null values.

camposandro: Good point! I set the `to_numpy()` "na_value" to our default and added a [wrapper](https://github.com/astronomy-commons/lsdb/blob/73df35f03659b26f2be64f5a9a4357386105aa3b/src/lsdb/catalog/dataset/healpix_dataset.py#L302C13-L302C25) for when we do not specify the skymap target order to also make sure we replace any null-value introduced by the user-defined function.

smcguire-cmu (approved):

Issue #89, opened 2023-12-12
Catalog loading may spend a lot of time on to_pyarrow_string


hombit:
When loading HiPSCat catalogs (for example PS1 DR2 OTMO), I noticed that LSDB spends up to ~50% of time on `to_pyarrow_string()`.

I think that the issue could be because current pipeline is like the following:
- Loading individual parquet file with pandas, using the default `numpy` dtype backend, here we convert `pyarrow` strings to `numpy`
- After we put it to Dask, and it converts strings back to `pyarrow`

hombit:
`read_hipscat(dtype_backend='pyarrow')` helps a lot. However I don't see an easy option to have all columns but string to be `numpy`.

See [this benchmark NB](https://github.com/lincc-frameworks/notebooks_lf/blob/d59a27d7c6a6df1ae0932ac51247c1e85e654f35/read-str-columns-with-lsdb.ipynb)

erykoff:
There is no easy way, that is true.  But pandas string handling is truly terrible, and going through that is definitely going to be a bottleneck.  (numpy string handling is terrible in a different way, and python string handling is terrible in yet a different way of course).

This was what I wrote for going parquet through arrow to numpy/astropy: https://github.com/astropy/astropy/blob/main/astropy/io/misc/parquet.py .  It has served me well, but I haven't actually checked how much the special casing of the strings slows it down.

hombit:
@erykoff thank you! I believe that here we have one more layer of the problem, Dask. It is Dask who converts string Series to `string[pyarrow]`. So I believe we have back and forth conversion happening in our pipeline: from parquet to numpy with `pd.read_parquet` and from numpy to arrow implicitly with Dask.

Personally I would prefer to slowly move to `pyarrow` dtypes since pandas v2 got some support of them.

Issue #303, opened 2024-05-02
Convert column dtypes to pyarrow by default in .from_dataframe()


hombit:
**Feature request**

As a part of our transition to pyarrow dtypes, it would be great to convert dtypes of input arrays in `.from_dataframe()`. We still may have an opt-out options, when users have pyarrow-incompatible types (i.e. `object`) or wish to keep their types for a different reason.

**Implementation ideas**

Dataframe types may be converted in this way:
```python
import pandas as pd
import pyarrow as pa

def convert_dtypes_to_pyarrow(df):
    new_series = {}
    for column in df.columns:
        try:
            pa_array = pa.array(df[column], from_pandas=True)
        except (ValueError, TypeError):
            raise ...
        series = pd.Series(
            pa_array,
            dtype=pd.ArrowDtype(pa_array.type),
            copy=False,
        )
        new_series[column] = series
    return pd.DataFrame(new_series, index=df.index, name=df.name, copy=False)
```

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [ ] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [x] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

PR #306, opened 2024-05-06
Load data using pyarrow types

### LINKED ISSUE #279: Support loading strings with the pyarrow backend to use the `string[pyarrow]` pandas dtype
Currently for string columns, pandas will load the strings as native python strings, and dask will then create a slow task to convert them all to pyarrow strings. Pandas has recently introduced support for the pyarrow string dtype, and can load strings from parquet files directly into a pandas df with the pyarrow string type by specifying `dtype_backend="pyarrow"` as an option in the `pd.read_parquet` call. 

We support passing kwargs to this function, but when generating the dask meta DataFrame from the parquet schema, we don't use pyarrow string types, and so we get a meta mismatch. So this needs to be updated here, and tested that the new dtype works with the other `from_delayed` functions for operations like crossmatching and joining where we generate the meta.

camposandro:
Adds the `use_pyarrow_types` argument to the `read_hipscat` interface. This allows us to pass the `dtype_backend` to the calls that read the parquet leaf files, as well as metadata, loading the data and respective schema with pyarrow types. 

Using the pyarrow backend (which is now the default!) we can load strings directly into a pandas dataframe with the pyarrow string type, avoiding the creation of slow tasks to convert python strings into pyarrow strings.

This change required me to also address #303, allowing catalogs to be created with pyarrow types in `from_dataframe`.

Closes #279 and #303.


hombit:
Does it also fix #89?

camposandro:
> Does it also fix #89?

Yes, the string conversion tasks should not be required for catalogs loaded with pyarrow types. 

I created a [notebook](https://github.com/lincc-frameworks/notebooks_lf/blob/main/sprints/2024/05_02/pyarrow-types/loading-catalogs-pyarrow-types.ipynb) that loads a chunk of ZTF sources with the "band" information, of string-type (we saw several bottlenecks happening with this column before). Using a single worker and the latest version of Dask (2024.5.0) these were the compute times I obtained:

**Previously (w/ pyarrow string conversion):** 181.35 s
**Previously (w/o pyarrow string conversion):** 88.31 s
**Currently (w/ pyarrow typed catalogs):** 83.81 s

src/lsdb/loaders/dataframe/from_dataframe_utils.py
```diff
@@ -12,22 +13,68 @@
...
+        ddf (dd.DataFrame): A Dask DataFrame
+
+    Returns:
+        A new dask DataFrame, where columns, index and schema have been
+        converted to use pyarrow types.
+    """
+    # Convert schema types according to the backend
+    pyarrow_meta = _convert_dtypes_to_pyarrow(ddf._meta)
+    # Apply the new schema to the dask dataframe
+    ddf = ddf.astype(pyarrow_meta.dtypes)
```
smcguire-cmu, line 51: Can we do this conversion before making it a ddf?

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -46,13 +47,16 @@ def __init__(
             highest_order (int): The highest partition order
             partition_size (int): The desired partition size, in number of rows
             threshold (int): The maximum number of data points per pixel
+            use_pyarrow_types (bool): If True, the data is backed by pyarrow, otherwise we keep the
+                original data types. Defaults to "pyarrow".
```
delucchi-cmu, line 51:
```suggestion
                original data types. Defaults to True.
```

src/lsdb/loaders/dataframe/from_dataframe.py
```diff
@@ -27,6 +28,8 @@ def from_dataframe(
         threshold (int): The maximum number of data points per pixel
         margin_order (int): The order at which to generate the margin cache
         margin_threshold (float): The size of the margin cache boundary, in arcseconds
+        use_pyarrow_types (bool): If True, the data is backed by pyarrow, otherwise we keep the
+            original data types. Defaults to "pyarrow".
```
delucchi-cmu, line 32:
```suggestion
            original data types. Defaults to True.
```

delucchi-cmu (approved):

smcguire-cmu (approved):

PR #322, opened 2024-05-14
Use MOCs for alignment and generate MOC in from_dataframe

smcguire-cmu:
Uses the MOC alignment functions recently added to hipscat to perform alignment for crossmatch and join with higher order coverage. Generates a MOC when `from_dataframe` is called.

### DIFF
--- a/src/lsdb/dask/merge_catalog_functions.py
+++ b/src/lsdb/dask/merge_catalog_functions.py
@@ -79,8 +79,12 @@ def align_catalogs(left: Catalog, right: Catalog, right_added_radius_arcsec: flo
     else:
         right_tree = right.hc_structure.pixel_tree
 
-    right_moc = right.hc_structure.moc
-    if right_moc is not None and right_added_radius_arcsec is not None:
+    right_moc = (
+        right.hc_structure.moc
+        if right.hc_structure.moc is not None
+        else right.hc_structure.pixel_tree.to_moc()
+    )
+    if right_added_radius_arcsec is not None:
         right_moc_depth_resol = hp.nside2resol(hp.order2nside(right_moc.max_order), arcmin=True) * 60
         if right_added_radius_arcsec < right_moc_depth_resol:
             right_moc = right_moc.add_neighbours()
--- a/src/lsdb/dask/merge_catalog_functions.py
+++ b/src/lsdb/dask/merge_catalog_functions.py
@@ -62,8 +62,8 @@ def align_catalogs(left: Catalog, right: Catalog, add_right_margin: bool = True)
     Args:
         left (lsdb.Catalog): The left catalog to align
         right (lsdb.Catalog): The right catalog to align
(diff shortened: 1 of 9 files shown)


src/lsdb/dask/merge_catalog_functions.py
```diff
@@ -54,25 +56,51 @@ def concat_partition_and_margin(
...
 
-def align_catalogs(left: Catalog, right: Catalog) -> PixelAlignment:
+def align_catalogs(left: Catalog, right: Catalog, add_right_margin: bool = True) -> PixelAlignment:
     """Aligns two catalogs, also using the right catalog's margin if it exists
 
     Args:
         left (lsdb.Catalog): The left catalog to align
         right (lsdb.Catalog): The right catalog to align
+        add_right_margin (float): When using MOCs to align catalogs, adds a border to the
+            right catalog's moc to include the margin of the right catalog, if the right margin exists
```
camposandro, line 66:
```suggestion
        add_right_margin (bool): If True, when using MOCs to align catalogs, adds a border to the
            right catalog's moc to include the margin of the right catalog, if it exists. Defaults to True.
```

src/lsdb/dask/merge_catalog_functions.py
```diff
@@ -54,25 +56,51 @@ def concat_partition_and_margin(
...
+        if right.hc_structure.moc is not None
+        else right.hc_structure.pixel_tree.to_moc()
+    )
+    if right_added_radius is not None:
+        right_moc_depth_resol = hp.nside2resol(hp.order2nside(right_moc.max_order), arcmin=True) * 60
+        if right_added_radius < right_moc_depth_resol:
+            right_moc = right_moc.add_neighbours()
+        else:
+            delta_order = int(np.ceil(np.log2(right_added_radius / right_moc_depth_resol)))
+            right_moc = right_moc.degrade_to_order(right_moc.max_order - delta_order).add_neighbours()
```
camposandro, line 95: Could we move this auxiliary code to hipscat?

camposandro (commented):
This looks awesome!

camposandro:
Could you also run a join using MOCs for one of those cases where not specifying a high lower order on `from_dataframe` would just render the workflow unfeasible? It'd be great to see how this change speeds up the computation. If I recall correctly [this](https://github.com/lincc-frameworks/notebooks_lf/blob/main/DELVE_gaia_xmatch/Spectroscopy.ipynb) was one of the workflows running into issues!

src/lsdb/dask/merge_catalog_functions.py
smcguire-cmu: Problem is hipscat catalog doesn't have the margin as a parameter. We should refactor a lot more stuff into hipscat I think, so let's open up a ticket to do that.

smcguire-cmu: opened #325

camposandro (approved):

PR #327, opened 2024-05-15
Recreate output directory on `to_hipscat` when using overwrite

### LINKED ISSUE #326: Saving catalog with overwrite should recreate the output directory
**Bug report**

When saving a catalog to disk using `to_hipscat` the provided output path may have content (e.g. if previously a catalog was saved to that same directory). In this case we need to specify the _overwrite_ flag to True.

However, the current behavior does not ensure that the output folder is erased and recreated. It only ensures that the previous catalog files that coincide with those of the new catalog are overwritten (parquet leaf files, as well as metadata). The parquet leaf files that do not directly match will the ones on the catalog currently being saved remain in that directory. The catalog saved to disk in this case is not correct.

We should throw an error alerting the user for the usage of the `overwrite` flag if it `overwrite` is False and the output directory has contents. If `overwrite` is True, the directory should be recreated from scratch.

**Before submitting**
Please check the following:

- [X] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.
- [X] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

camposandro:
Saving a catalog to disk with `to_hipscat` and _overwrite_ set to True replaces the previous files that coincide (in naming) to those of the new catalog but those that don't remain in that same directory. This pull request changes this behavior, recreating the output directory entirely if it has contents and the `overwrite` flag is set to True, and issuing an error if it has contents and `overwrite` is set to False. Closes #326.


src/lsdb/io/to_hipscat.py
```diff
@@ -70,13 +70,20 @@ def to_hipscat(
...
+                f"base_catalog_path ({base_catalog_dir_fp}) contains files."
+                " choose a different directory or set overwrite to True."
+            )
+        hc.io.file_io.remove_directory(base_catalog_dir_fp, storage_options=storage_options)
+    hc.io.file_io.make_directory(base_catalog_dir_fp, exist_ok=True, storage_options=storage_options)
     # Save partition parquet files
     pixel_to_partition_size_map = write_partitions(catalog, base_catalog_dir_fp, storage_options, **kwargs)
     # Save parquet metadata
-    hc.io.write_parquet_metadata(base_catalog_path, storage_options, **kwargs)
+    hc.io.write_parquet_metadata(base_catalog_path, storage_options)
```
smcguire-cmu, line 86: why are the kwargs removed?

smcguire-cmu (approved):

PR #328, opened 2024-05-20
Allow margin path argument in `read_hipscat`

### LINKED ISSUE #314: support string path for loading margin with `read_hipscat(margin_cache="...")`
Currently to load a margin you need two read_hispcat calls for the margin first then specify the loaded margin catalog in the catalog read. We should allow the option to simplify this by specifying the margin cache as a path to load.

camposandro:
Adds a `margin_path` argument to the `read_hipscat` interface allowing for standard HiPSCat catalogs to be loaded with margins (in parquet format) from their path on disk. Closes #314.

### DIFF
--- a/src/lsdb/loaders/hipscat/hipscat_catalog_loader.py
+++ b/src/lsdb/loaders/hipscat/hipscat_catalog_loader.py
@@ -1,8 +1,11 @@
+from __future__ import annotations
+
 import dataclasses
 
 import hipscat as hc
 
-from lsdb.catalog.catalog import Catalog
+from lsdb.catalog.catalog import Catalog, MarginCatalog
+from lsdb.loaders import read_hipscat
 from lsdb.loaders.hipscat.abstract_catalog_loader import AbstractCatalogLoader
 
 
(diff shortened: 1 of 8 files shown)


src/lsdb/loaders/hipscat/hipscat_catalog_loader.py
```diff
@@ -34,3 +37,20 @@ def _filter_hipscat_catalog(self, hc_catalog: hc.catalog.Catalog) -> hc.catalog.
...
+        margin_catalog = None
+        if self.config.margin_cache is not None:
+            margin_catalog = self.config.margin_cache
+        elif self.config.margin_path is not None:
+            margin_catalog = lsdb.read_hipscat(
+                path=self.config.margin_path,
+                catalog_type=MarginCatalog,
+                search_filter=self.config.search_filter,
+                dtype_backend=self.config.dtype_backend,
+                storage_options=self.storage_options,
```
smcguire-cmu, line 53: add `margin_path = None` pls

src/lsdb/loaders/hipscat/hipscat_loading_config.py
```diff
@@ -26,21 +26,22 @@ class HipscatLoadingConfig:
     margin_cache: MarginCatalog | None = None
     """Margin cache for the catalog. By default, it is None"""
 
+    margin_path: str | None = None
+    """The path for the margin cache catalog. By default, it is None"""
+
```
smcguire-cmu, line 31: Can these be one argument and check the type?

smcguire-cmu (approved):
Approved, but don't merge until tomorrow after the demos

PR #334, opened 2024-05-23
Add catalog partition indexer

### LINKED ISSUE #131: Mirror dask partiton in catalog object to see subset of the data
Enable user to see a subset of the catalog (in an individual partition), in a similar way as dask.ddf.partitions

camposandro:
Implements an indexer to allow slicing of catalog partitions using the square brackets accessor. The accessor accepts a single integer or a list of integers (e.g. `catalog.partitions[2]` returns a new catalog with only the third partition; `catalog.partitions[::2]` returns a new catalog with all the even-indexed partitions). Closes #131.


src/lsdb/core/search/pixel_search.py
```diff
@@ -0,0 +1,26 @@
...
+
+    Filters partitions in the catalog to those that are in a specified pixel set.
+    Does not filter points inside those partitions.
+    """
+
+    def __init__(self, pixels):
+        self.pixels = set(pixels)
+
+    def search_partitions(self, pixels: List[HealpixPixel]) -> List[HealpixPixel]:
+        return [p for p in pixels if p in self.pixels]
```
delucchi-cmu, line 23: Could we do pixel overlap instead of exact match?

For the use case of this PR, you want exact match, but I would love something like:

`read_hipscat("path", filter=PixelSearch([(3, 34)])`

to limit the catalog to a big chunk of the sky (that's order 3, so would likely contain lots of partitions inside it).

delucchi-cmu (approved):

Issue #308, opened 2024-05-07
read_hipscat(search_filter) is confusing


hombit:
**Bug report**

It is not clear from the API docs (and there are no RTD docs) that `read_hipscat(search_filter=...)` would not filter the data, but only the partitions. It is also confusing that `Catalog._search` is private, so user is not supposed to reuse the same search object: `lsdb.read_hipscat(search_filter=search).search(search)`.

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [ ] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.
- [ ] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

nevencaplar:
I agree. I believe we did this ``coarse'' so that it is faster, but it is very unclear and misleading.

delucchi-cmu:
Sandro - can you write up some alternatives for discussion?

camposandro:
We decided to change the current behavior of `read_hipscat(search_filter)` to meet the users' expectations, performing the fine search in addition to the partition coarse search. We should try to add `fine` as a property of the search object and make the internal [`_search`](https://github.com/astronomy-commons/lsdb/blob/0ab3345ec3ce501d9c892443c6f4f83a3f84c32c/src/lsdb/catalog/catalog.py#L299) method public so users can reuse their search objects if desired.

Issue #129, opened 2024-01-26
Suggest improvements to API names


nevencaplar:
Investigate the names and conventions in astronomical software packages; ensure that our names are consistent and as clear as possible. As an example, I find that `.read_hipscat()` is somewhat confusing as ``read'' part implies that something is actually being done, and it not consistent with lazy load paradigm. 

The end result is a set of suggestion to change the name of certain methods. 

Implementation to be described in a separate ticket.

camposandro:
LSDB, unlike other astronomical packages, operates lazily by nature. Less than a handful of methods implicitly run computations for the user (e.g. `head`, `prune_empty_partitions`, `skymap`), so we decided to keep `read_hipscat`'s current naming and provide more disclaimers in the documentation about this behavior. 

The `*_search` catalog methods are coherently named, and follow the astropy / astropy-healpix decision of having the search keyword in the signature. The exception is `box`, which we should rename to `box_search`.

camposandro:
Closed by #345. A diagram of the LSDB pipeline should further clarify the lazy nature of the package #332.

PR #345, opened 2024-06-04
Rename search method and add fine as a search object argument

camposandro:
This small pull request:
- Changes `box` to `box_search` to ensure naming consistency;
- Changes the catalog's "search" method visibility to public;
- Moves the fine filtering argument to a search object attribute;

Targets part of issues #129 and #308.


delucchi-cmu (approved):
Great changes!

PR #350, opened 2024-06-06
Adds fine filtering on `read_hipscat`

### LINKED ISSUE #308: read_hipscat(search_filter) is confusing

camposandro:
Adds fine filtering to the `read_hipscat` operation. Previously, when provided with a `search_filter`, the catalog was loaded with the pixels within the specified spatial region but we did not filter their data. This behavior created confusion with the users so we decided to apply the full search. We still do it in two steps, first figuring out which subset of files we should read from disk, creating the Dask DataFrame with the partitions, and then performing filtering on each of them. Closes #308.


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -130,22 +130,21 @@ def query(self, expr: str) -> Self:
...
+        metadata: hc.catalog.Catalog | hc.catalog.MarginCatalog,
         search: AbstractSearch,
-    ):
+    ) -> Tuple[dict, dd.core.DataFrame]:
         """Performs a search on the catalog from a list of pixels to search in
 
         Args:
-            metadata (hc.catalog.Catalog): The metadata of the hipscat catalog.
-            filtered_pixels (List[HealpixPixel]): List of pixels in the catalog to be searched.
+            metadata (hc.catalog.Catalog | hc.catalog.MarginCatalog): The metadata of the hipscat catalog.
```
smcguire-cmu, line 139: Can you add that this is the metadata of the catalog after filtering

src/lsdb/loaders/hipscat/abstract_catalog_loader.py
```diff
@@ -50,7 +46,7 @@ def _load_hipscat_catalog(self, catalog_type: Type[HCCatalogTypeVar]) -> HCCatal
         """Load `hipscat` library catalog object with catalog metadata and partition data"""
         return catalog_type.read_from_hipscat(self.path, storage_options=self.storage_options)
 
-    def _load_dask_df_and_map(self, catalog: HCHealpixDataset) -> Tuple[dd.core.DataFrame, DaskDFPixelMap]:
+    def _load_dask_df_and_map(self, catalog: HCCatalogTypeVar) -> Tuple[dd.core.DataFrame, DaskDFPixelMap]:
```
smcguire-cmu, line 49: Why is this the type var here?

smcguire-cmu: And everywhere else in the file

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -136,7 +136,9 @@ def _perform_search(
         """Performs a search on the catalog from a list of pixels to search in
 
         Args:
-            metadata (hc.catalog.Catalog | hc.catalog.MarginCatalog): The metadata of the hipscat catalog.
+            metadata (hc.catalog.Catalog | hc.catalog.MarginCatalog): The metadata of
```
smcguire-cmu, line 139: Could this be `HealpixCatalog` instead of `Catalog | MarginCatalog`?

src/lsdb/types.py
```diff
@@ -1,19 +1,11 @@
...
-
 # Compute pixel map returns a tuple. The first element is
 # the number of data points within the HEALPix pixel, the
 # second element is the list of pixels it contains.
 HealpixInfo: TypeAlias = Tuple[int, List[int]]
-DaskDFPixelMap = Dict[HealpixPixel, int]
 
-# Generic lsdb and hipscat catalog types
-CatalogTypeVar = TypeVar("CatalogTypeVar", bound=Dataset)
-HCCatalogTypeVar = TypeVar("HCCatalogTypeVar", bound=HCHealpixDataset)
```
smcguire-cmu: If it works, I actually like having the type vars in this file

src/lsdb/catalog/dataset/healpix_dataset.py
camposandro: Unfortunately not, the HealpixDataset's `BaseCatalogInfo` does not have the ra/dec information required for filtering, only these subclasses do; mypy complains.

smcguire-cmu (approved):

PR #353, opened 2024-06-10
Allow path-like objects in the margin cache argument

### LINKED ISSUE #343: read_hipscat(margin_cache=...) doesn't support Path objects
**Bug report**

It would be great if `margin_cache` argument of `read_hipscat` would support all the same object types as the first positional argument, in addition to `Catalog` instances. This would include `pathlib.Path` objects and, probably, `fsspec` objects.

```python
from lsdb import read_hipscat

read_hipscat(Path(...))  # works
read_hipscat("...", margin_cache="...")  # works
read_hipscat(Path(...), margin_cache=read_hipscat(Path(...)))  # works
read_hipscat(Path(...), margin_cache=Path(...))  # fails
```

The possible solution is to check first if the argument type is `Catalog` and use it as is if not, without further type checking

**Before submitting**
Please check the following:

- [  ] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

camposandro:
Removes the overly strict check on the margin cache argument to allow the use of path-like objects. Closes #343.

### DIFF
--- a/tests/lsdb/loaders/hipscat/test_read_hipscat.py
+++ b/tests/lsdb/loaders/hipscat/test_read_hipscat.py
@@ -94,11 +94,6 @@ def test_catalog_without_margin_is_none(small_sky_xmatch_dir):
     assert catalog.margin is None
 
 
-def test_catalog_with_wrong_margin_args(small_sky_xmatch_dir):
-    with pytest.raises(ValueError, match="must be of type"):
-        lsdb.read_hipscat(small_sky_xmatch_dir, margin_cache=1)
-
-
 def test_read_hipscat_subset_with_cone_search(small_sky_order1_dir, small_sky_order1_catalog):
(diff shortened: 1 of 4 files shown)


tests/lsdb/loaders/hipscat/test_read_hipscat.py
```diff
@@ -94,11 +99,6 @@ def test_catalog_without_margin_is_none(small_sky_xmatch_dir):
     assert catalog.margin is None
 
 
-def test_catalog_with_wrong_margin_args(small_sky_xmatch_dir):
-    with pytest.raises(ValueError, match="must be of type"):
-        lsdb.read_hipscat(small_sky_xmatch_dir, margin_cache=1)
-
-
```
hombit, line 101: Why do you delete this one?

camposandro: I removed the type checking so this error isn't thrown.

hombit (approved):

PR #358, opened 2024-06-13
Specify argument name for optional storage_options argument.

delucchi-cmu:
See cloudtests smoke failure when using `storage_options`: https://github.com/astronomy-commons/hipscat-cloudtests/actions/runs/9495013490


dougbrn (approved):
looks good!

PR #364, opened 2024-06-21
Change to using dask expressions

smcguire-cmu:
Updates dask DataFrame calls to use expressions methods instead of legacy methods. Updates DataFrame creation methods to check for empty delayed lists which now fail with dask expressions. Updates search to use map_partitions instead of delayed for better performance.

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -154,7 +154,7 @@ class HealpixDataset(Dataset):
             filtered_partitions_ddf = filtered_partitions_ddf.map_partitions(
                 search.search_points,
                 metadata.catalog_info,
-                meta=filtered_partitions_ddf,
+                meta=self._ddf._meta,
                 transform_divisions=False,
             )
         ddf_partition_map = {pixel: i for i, pixel in enumerate(filtered_pixels)}
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -152,7 +152,10 @@ class HealpixDataset(Dataset):
         filtered_partitions_ddf = self._ddf.partitions[target_partitions_indices]
         if search.fine:
             filtered_partitions_ddf = filtered_partitions_ddf.map_partitions(
-                search.search_points, metadata, meta=filtered_partitions_ddf, transform_divisions=False
+                search.search_points,
+                metadata.catalog_info,
+                meta=filtered_partitions_ddf,
+                transform_divisions=False,
             )
         ddf_partition_map = {pixel: i for i, pixel in enumerate(filtered_pixels)}
         return ddf_partition_map, filtered_partitions_ddf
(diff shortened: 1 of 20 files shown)


src/lsdb/loaders/dataframe/margin_catalog_generator.py
```diff
@@ -73,37 +73,39 @@ def create_catalog(self) -> MarginCatalog | None:
...
+        if len(pixels) == 0:
             return None
+        ddf, ddf_pixel_map, total_rows = self._generate_dask_df_and_map(pixels, partitions)
+        margin_pixels = list(ddf_pixel_map.keys())
         margin_catalog_info = self._create_catalog_info(total_rows)
         margin_structure = hc.catalog.MarginCatalog(margin_catalog_info, margin_pixels)
         return MarginCatalog(ddf, ddf_pixel_map, margin_structure)
 
-    def _generate_dask_df_and_map(self) -> Tuple[dd.core.DataFrame, Dict[HealpixPixel, int], int]:
+    def _get_margins(self):
```
camposandro, line 85:
```suggestion
    def _get_margins(self) -> Tuple[List[HealpixPixel], List[pd.DataFrame]]:
```

Could you add a brief docstring too?

src/lsdb/loaders/dataframe/margin_catalog_generator.py
```diff
@@ -73,37 +73,39 @@ def create_catalog(self) -> MarginCatalog | None:
...
+        combined_pixels = (
+            self.hc_structure.get_healpix_pixels() + self.hc_structure.generate_negative_tree_pixels()
+        )
+        margin_pairs_df = self._find_margin_pixel_pairs(combined_pixels)
+        margins_pixel_df = self._create_margins(margin_pairs_df)
+        pixels, partitions = list(margins_pixel_df.keys()), list(margins_pixel_df.values())
+        return pixels, partitions
+
+    def _generate_dask_df_and_map(
+        self, pixels, partitions
```
camposandro, line 95:
```suggestion
        self, pixels: List[HealpixPixel], partitions: List[pd.DataFrame]
```

Can we update the docstring as well?

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -167,9 +172,12 @@ def _construct_search_ddf(
         Returns:
             The catalog pixel map and the respective Dask DataFrame
         """
+        filtered_partitions = (
+            filtered_partitions if len(filtered_partitions) > 0 else [delayed(self._ddf._meta)]
+        )
         divisions = get_pixels_divisions(filtered_pixels)
-        search_ddf = dd.io.from_delayed(filtered_partitions, meta=self._ddf._meta, divisions=divisions)
-        search_ddf = cast(dd.core.DataFrame, search_ddf)
+        search_ddf = dd.from_delayed(filtered_partitions, meta=self._ddf._meta, divisions=divisions)
```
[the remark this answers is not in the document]
smcguire-cmu: yeah good point, doesn't even need a map_partitions call just a filter using partitions, so I changed it

camposandro (approved):

Issue #377, opened 2024-07-18
from_dataframe improvements


delucchi-cmu:
As discussed in design meeting on July 18:

- [ ] `partition_size` argument should be the max number of bytes (i.e. in-memory size) for each partition. this mimics [dask dataframe](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.repartition.html) behavior, and distinguishes it from the `threshold` parameter
- [ ] add a warning when input dataframe exceeds 1m rows OR 1G in-memory size. warning message should point to either the documentation generated in [this issue](https://github.com/astronomy-commons/lsdb/issues/376), or the hipscat-import RTD (or both)
- [ ] default values for lowest order, highest order, and threshold should be identical for `from_dataframe` and hipscat-import catalog import arguments. i know that i requested these be changed prior to the MOC improvements, and i'm sorry for the reversal.

Issue #376, opened 2024-07-18
Documentation: highlight that from_dataframe is not for large datasets


hombit:
We should highlight that `from_dataframe()` is not designed for large datasets (>~1M rows) and that `hipscat-import` should be used instead. This should go to both docstring and RTD

smcguire-cmu:
This would also be good as a warning when we run `from_dataframe` to check the size of the input data frame and raise a warning if it's large.

camposandro:
This has been completed by #379.

PR #378, opened 2024-07-18
Docs hot fixes

delucchi-cmu:
Related to https://github.com/astronomy-commons/lsdb/issues/376

- use data.lsdb.io instead of epyc, where possible.
- expand on reasons for using from_dataframe vs hipscat-import, and provide link to that package's RTD
- don't do extra install of aiohttp (see https://github.com/astronomy-commons/hipscat/pull/291)

nb: we're not ready to take the `lowest_healpix_order` references out of the "import catalogs" notebook yet, as the two modes have different defaults and won't result in the same catalog. see https://github.com/astronomy-commons/lsdb/issues/377

### DIFF
--- a/docs/tutorials/import_catalogs.ipynb
+++ b/docs/tutorials/import_catalogs.ipynb
@@ -9,7 +9,10 @@
    "source": [
     "# Importing catalogs to HiPSCat format\n",
     "\n",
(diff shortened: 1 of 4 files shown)


delucchi-cmu:
See docs preview on RTD: https://lsdb--378.org.readthedocs.build/en/378/tutorials/import_catalogs.html

nevencaplar (approved):
Looks good. Thank you!

PR #383, opened 2024-07-24
Provide arrow schema on HiPSCat catalog creation

camposandro:
Add the arrow schema to operations that require instantiating new HiPSCat catalogs.

### DIFF
--- a/src/lsdb/io/schema.py
+++ b/src/lsdb/io/schema.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import List
-
 import dask.dataframe as dd
 import pyarrow as pa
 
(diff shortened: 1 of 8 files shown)


src/lsdb/loaders/hipscat/hipscat_catalog_loader.py
```diff
@@ -53,13 +55,15 @@ def _load_margin_catalog(self) -> MarginCatalog | None:
...
-                dtype_backend=self.config.dtype_backend,
+            margin_catalog = MarginCatalogLoader(
+                str(self.config.margin_cache),
+                HipscatLoadingConfig(
+                    search_filter=self.config.search_filter,
+                    columns=self.config.columns,
+                    margin_cache=None,
+                    dtype_backend=self.config.dtype_backend,
+                    **self.config.kwargs,
+                ),
```
camposandro, line 66: This change is due to a circular import issue reported by pylint.

delucchi-cmu (approved):

delucchi-cmu (approved):
We may want to expand our implementation of `get_arrow_schema` in the future, to do more manual manipulation of the schema, since pandas doesn't always get the arrow schema right. But this looks great for now!

camposandro:
Yes, Pandas's behavior might be a bit unpredictable. For now the Dask meta for the catalogs is generated from the arrow schema at loading time so using that same meta for schema merges works well. But I agree, especially for joining and crossmatching, it'd be great to have that manual generation. I'll open an issue and link it to this PR, thanks Melissa!

Issue #372, opened 2024-07-15
LSDB with Nested-Dask: Additional Wrappers


dougbrn:
**Feature request**
Related to #367. Add wrappers for Nested-Dask functions directly to the catalog API. The functions that should have wrappers are:
* dropna
* reduce

**Before submitting**
Please check the following:

- [ ] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [ ] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [ ] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

smcguire-cmu:
Closed with #414 and #410

PR #410, opened 2024-08-26
Wrap nested_dask dropna function

smcguire-cmu:
Part of #372, wraps the dropna function from nested_dask as a catalog method

### DIFF
--- a/tests/lsdb/catalog/test_nested.py
+++ b/tests/lsdb/catalog/test_nested.py
@@ -1,5 +1,5 @@
-import numpy as np
 import nested_dask as nd
+import numpy as np
 import pandas as pd
 
 from lsdb import Catalog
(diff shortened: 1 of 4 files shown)


tests/lsdb/catalog/test_nested.py
```diff
@@ -0,0 +1,15 @@
+import nested_dask as nd
+import pandas as pd
+
+from lsdb import Catalog
+
+
+def test_dropna(small_sky_with_nested_sources):
+    filtered_cat = small_sky_with_nested_sources.query("sources.mag < 15.1")
+    drop_na_cat = filtered_cat.dropna()
```
dougbrn, line 9: It might be good to also add a test that checks the unique nested-dask dropna behavior, so something that would use the subset or on_nested kwargs

dougbrn (approved):
Looks good! Thanks Sean. Just recommending you add an additional test.

dougbrn (approved):
Thanks Sean, looks good!

PR #413, opened 2024-09-12
Replace FilePointer with universal pathlib

delucchi-cmu:
See https://github.com/astronomy-commons/hipscat/pull/336

## Solution Description

Use fsspec's project, [universal-pathlib](https://github.com/fsspec/universal_pathlib) as a path provider. In this way, everywhere we currently pass in a `FilePointer` we can use a `Path` instead, and when we need to create remote connections, use `UPath`.


delucchi-cmu:
Documentation issue will require https://github.com/astronomy-commons/hipscat-import/pull/387/, but I'd like to send for review for general comments.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -408,10 +409,9 @@ def plot_pixels(self, projection: str = "moll", **kwargs):
 
     def to_hipscat(
         self,
-        base_catalog_path: str,
+        base_catalog_path: UPath,
```
smcguire-cmu, line 412: Can we have the user facing functions as `UPath | str`? Like this and `read_hipscat`

src/lsdb/loaders/hipscat/abstract_catalog_loader.py
```diff
@@ -105,3 +93,12 @@ def _get_kwargs(self) -> dict:
         if self.config.dtype_backend is not None:
             kwargs["dtype_backend"] = self.config.dtype_backend
         return kwargs
+
+
+def read_pixel(
```
smcguire-cmu, line 98: Can we add a specific columns kwarg to this? The way `from_map` with dask expressions works is that if the method has a columns param then it use it for query optimization, but if not will just always load all columns.

smcguire-cmu (commented):
Soo nice to be getting rid of storage options from everywhere

src/lsdb/catalog/dataset/healpix_dataset.py
delucchi-cmu: Switched a bunch of places to use `str | Path | UPath`.

smcguire-cmu (approved):

PR #422, opened 2024-09-26
Initial work toward properties file.

delucchi-cmu:
Mostly a drop-in replacement!

- use copy/modify instead of dataclasses.replace
- refreshes the unit test data
- removes provenance info writing
- update for newest pylint rules


src/lsdb/catalog/catalog.py
```diff
@@ -469,12 +465,11 @@ def join(
             if output_catalog_name is None:
                 output_catalog_name = self.hc_structure.catalog_info.catalog_name
 
-            new_catalog_info = dataclasses.replace(
-                self.hc_structure.catalog_info,
-                catalog_name=output_catalog_name,
-                ra_column=self.hc_structure.catalog_info.ra_column + suffixes[0],
-                dec_column=self.hc_structure.catalog_info.dec_column + suffixes[0],
-            )
+            new_catalog_info = self.hc_structure.catalog_info.model_copy()
```
gitosaurus, line 468:
```suggestion
            new_catalog_info = self.hc_structure.catalog_info.model_copy(update=dict(
                catalog_name=output_catalog_name,
                ra_column=catalog_info.ra_column + suffixes[0],
                dec_column=catalog_info.dec_column + suffixes[0],
            ))
```
If you want to maintain the same kind of keyword-argument initialization of a new instance with Pydantic that you had with dataclasses, the above works.

src/lsdb/io/to_hats.py
```diff
@@ -90,13 +88,9 @@ def to_hats(
...
     )
-    hc.io.write_catalog_info(catalog_base_dir=base_catalog_path, dataset_info=new_hc_structure.catalog_info)
+    new_hc_structure.catalog_info.to_properties_file(base_catalog_path)
     # Save provenance info
-    hc.io.write_metadata.write_provenance_info(
-        catalog_base_dir=base_catalog_path,
-        dataset_info=new_hc_structure.catalog_info,
-        tool_args=_get_provenance_info(new_hc_structure),
-    )
+    ## TODO - optionally write out arguments file
```
gitosaurus, line 93: Push to another issue?

src/lsdb/io/to_hats.py
```diff
@@ -176,34 +170,9 @@ def create_modified_catalog_structure(
...
-        **structure_args,
-        **dataclasses.asdict(catalog_structure.catalog_info),
-    }
-    provenance_info = {
-        "tool_name": "lsdb",
-        "version": version("lsdb"),
-        "runtime_args": args,
-    }
-    return provenance_info
+    new_hc_structure.catalog_info = new_hc_structure.catalog_info.model_copy()
```
gitosaurus, line 174:
```suggestion
    new_hc_structure.catalog_info = new_hc_structure.catalog_info.model_copy(update=dict(
        catalog_name=catalog_name,
        **kwargs
    ))
```
In this case, I do think the above pattern is worth preferring to the loop with `setattr`.

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -112,20 +110,32 @@ def _calculate_threshold(self, partition_size: int | None = None, threshold: int
         return threshold
 
     @staticmethod
-    def _create_catalog_info(**kwargs) -> CatalogInfo:
+    def _create_catalog_info(
+        catalog_name: str = "from_lsdb_dataframe",
+        ra_column: str = "ra",
+        dec_column: str = "dec",
+        catalog_type=CatalogType.OBJECT,
```
gitosaurus, line 117:
```suggestion
        catalog_type: CatalogType = CatalogType.OBJECT,
```

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -112,20 +110,32 @@ def _calculate_threshold(self, partition_size: int | None = None, threshold: int
...
+        **kwargs,
+    ) -> TableProperties:
         """Creates the catalog info object
 
         Args:
+            catalog_name: it is recommended to provide a new name for your catalog
+            ra_column: column to find right ascension coordinate
+            dec_column: column to find declination coordinate
+            catalog_type: type of table being created (e.g. object, margin, index)
```
gitosaurus, line 126:
```suggestion
            catalog_type: type of table being created (e.g. OBJECT, MARGIN, INDEX)
```
To signal that these are enumeration values.  I do see that the enumeration is based on `str` and that the string values of those enumerations are indeed the lowercase names, so this is a very light suggestion.

gitosaurus (commented):
Everything looks correct, but consider the suggestions.

src/lsdb/catalog/catalog.py
gitosaurus: Ah, I'm seeing that my suggestion won't catch misspellings of field names in the way that `dataclasses.replace` does, so it's probably better to set individual fields the way you had been, in these cases.

src/lsdb/io/to_hats.py
gitosaurus: Although, as before, `setattr` will do the spell check that `update=` doesn't.

gitosaurus (approved):
Giving approval since my suggestions are optional, and in the `update=` case, may be worth ignoring.

src/lsdb/catalog/catalog.py
delucchi-cmu: With `extra=True` in pydantic, there's not a TON of hope of catching misspellings. But I like the idea of copying and setting new field values all in a single call!

pylint isn't happy with the `dict()` and would prefer `{}`.

src/lsdb/io/to_hats.py
delucchi-cmu: Removed the TODO. On further reflection, I don't think we have any additional fields we'd want to write to such a file.

src/lsdb/io/to_hats.py
delucchi-cmu: Dolphinately.

gitosaurus: > Dolphinately.

I was wrong about that: Pydantic models don't notice the misspellings *anyway*, unless `Config: extra = 'forbid'`, and it sounds like we don't do that.  (And even if we did, `update=` still wouldn't check.)

src/lsdb/catalog/catalog.py
gitosaurus: > With `extra=True` in pydantic, there's not a TON of hope of catching misspellings. But I like the idea of copying and setting new field values all in a single call!
> 
> pylint isn't happy with the `dict()` and would prefer `{}`.

I didn't realize that!  Poking around, it appears that this preference is driven by performance.

https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/use-dict-literal.html

PR #423, opened 2024-09-30
Fix reference to partition info constant

delucchi-cmu:
Constants of `PartitionInfo.METADATA_*` can be replaced with `paths.PARTITION_*`.

Replaces awkward `model_copy(update=...)` with convenience method `copy_and_update`, which performs explicit model validation.

Re-imports unit test data with recent changes to hats and hats-import.


src/lsdb/loaders/dataframe/margin_catalog_generator.py
```diff
@@ -206,20 +210,27 @@ def _get_data_in_margin(
         )
         return partition_df.iloc[margin_mask]
 
-    def _create_catalog_info(self, total_rows: int) -> TableProperties:
+    def _create_catalog_info(self, catalog_name: str | None = None, **kwargs) -> TableProperties:
         """Create the margin catalog info object
 
         Args:
-            total_rows (int): The number of elements in the margin catalog
+            **kwargs: Arguments to pass to the creation of the catalog info
```
gitosaurus, line 217: `total_rows` is gone, now, but should `catalog_name` be described in the args?

gitosaurus (approved):
Looks good!  I like the new `.copy_and_update` a lot.

PR #429, opened 2024-10-08
Override catalog's `__len__` method

### LINKED ISSUE #404: Add `len` support for unmodified catalogs
**Feature request**

We can get the length of a catalog after it has been loaded and before any filtering operations, which some users may find useful. We can use the `len` for this and return None or make a warning if a catalog has been filtered.

camposandro:
Override the built-in `__len__` method to provide the number of rows for unmodified catalogs. I also added docstrings on `query` and `dropna` for better clarity (they existed in the base class but not in `Catalog`). Closes #404.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-git+https://github.com/astronomy-commons/hipscat.git@issue/333/catalog-len-method
+git+https://github.com/astronomy-commons/hipscat.git@development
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -1 +1 @@
-git+https://github.com/astronomy-commons/hipscat.git@development
\ No newline at end of file
+git+https://github.com/astronomy-commons/hipscat.git@issue/333/catalog-len-method
```
smcguire-cmu, line 1: Can you change this after merging the hipscat pr?

smcguire-cmu (approved):

PR #438, opened 2024-10-14
Use Alignment moc in crossmatch/join operations

### LINKED ISSUE #428: MOCs are lost when performing join/crossmatch
Currently crossmatching/joining loses MOC info, since the aligned MOC isn't returned by the hipscat alignment functions. So after crossmatching or joining, the benefits of the MOCs are lost.

I think the best way to fix this is to include a `moc` attribute in the `PixelAlignment` class, and use that in the crossmatch/join catalog methods.

smcguire-cmu:
Uses the `PixelAlignment` object's moc as the moc in the output catalogs of join and crossmatch operations.

Fixes #428

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -10,4 +10,4 @@ sphinx-autoapi
 sphinx-copybutton
 sphinx-book-theme
 sphinx-design
-git+https://github.com/astronomy-commons/hipscat.git@main
+git+https://github.com/astronomy-commons/hipscat.git@development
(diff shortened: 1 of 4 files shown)


camposandro (commented):
I noticed that the filtered structures on search operations are also not carrying the mocs: 
https://github.com/astronomy-commons/lsdb/blob/c478401626ccd5a368987847003cf12e7c52c851/src/lsdb/catalog/catalog.py#L326

Do you want to address it in this PR or should I open a new issue?

smcguire-cmu:
I'm pretty sure the searches do carry the mocs. The `filter_hc_catalog` method calls the `filter_by_moc` or `filter_by_healpix_pixels` methods in hipscat that include the filtered moc.

camposandro (approved):

Issue #178, opened 2024-02-23
Inherited catalog methods should use `self.__class__()` to init returned objects


smcguire-cmu:
Currently, when creating new catalogs for returned objects, we use a specific class. This means in classes that inherit it, they will return the base class instead of the object's class. If instead, we use `self.__class__` it will make the same type of object.

PR #447, opened 2024-10-21
Fix shallow copy of catalog structure

camposandro:
Fix the catalog structure, which was being shallow copied, for operations such as `query`, `dropna` and `nest_lists`. It adds a method that performs the deep copy of the underlying structure and that allows to override catalog info parameters (e.g. we invalidate rows setting `total_rows` to 0). It further adds a missing docstring to `reduce`.


camposandro:
> Does this close #178?

Not quite, I think there's still some references that were left behind. But it does close #454, just added it to the PR!

PR #449, opened 2024-10-21
Always load index column with no pandas metadata

### LINKED ISSUE #446: Without pandas metadata, index column isn't loaded when user specifies columns to read_hats

smcguire-cmu:
Previously, if there was pandas metadata the index column would be always loaded and set as the DataFrame index, even if the users specified columns that didn't include it. With no pandas metadata, the index column would be ignored and the default pandas index used in the DataFrame.

This updates catalog loading to always load the index column if available, with or without pandas metadata.

Fixes #446


tests/conftest.py
```diff
@@ -75,6 +76,11 @@ def small_sky_order1_dir(test_data_dir):
     return test_data_dir / SMALL_SKY_ORDER1_DIR_NAME
 
 
+@pytest.fixture
+def small_sky_order1_no_pandas_dir(test_data_dir):
+    return test_data_dir / SMALL_SKY_ORDER1_NO_PANDAS_DIR_NAME
```
delucchi-cmu, line 81: Note that with https://github.com/astronomy-commons/hats-import/pull/415, this behavior will be the default.

delucchi-cmu (approved):

tests/conftest.py
smcguire-cmu: Got it! I think when that's published we can regenerate the lsdb test data and I'll change this to test with the pandas metadata

PR #457, opened 2024-10-23
Fix benchmarking data for HATS migration

delucchi-cmu:
Adds data and data-generation-notebook for the two metadata-only catalogs in the benchmarking directory.


camposandro (approved):
Thank you!!

PR #458, opened 2024-10-23
Disable fine filtering in margin generation for from_dataframe

### LINKED ISSUE #408: Speed-up margin cache generation in from_dataframe()
I propose to turn off fine filtering and use a more accurate margin order guessing algorithm when calling `from_dataframe()`. It would significantly speedup catalog generation introducing a small memory overhead.

Both changes were recently introduced in `hipscat-import`: we may adopt them:
https://github.com/astronomy-commons/hipscat-import/pull/356
https://github.com/astronomy-commons/hipscat-import/pull/366

delucchi-cmu:
Closes #408


camposandro (approved):
Thanks Melissa, looks good!

PR #461, opened 2024-10-25
Pre-factor to simplify hats loading

### LINKED ISSUE #460: No properties file found error should be more descriptive
If you try to load a catalog with `lsdb.read_hats` it fails with `FileNotFoundError: No properties file found where expected:`

I suggest to make the error more descriptive, such ``not able to find HATS catalog at specified location`` or similar.

delucchi-cmu:
Complements (and requires) https://github.com/astronomy-commons/hats/pull/394

Closes #460.

Puts all the file reading on loading a catalog into a single module. This reduces the complexity for reading the code, reduces code duplication, and also reduces the number of times we need to read the `properties` file, and perform file system `stat` calls to determine if files exist. This should speed up catalog loading on remote file systems where there is a latency associated with each file read.

Enough of the code has changed, that we no longer need the additional (and mostly unused) argument for catalog type, in order to appease mypy type checking.


src/lsdb/loaders/hats/read_hats.py
```diff
@@ -69,23 +60,169 @@ def read_hats(
...
-    return dataset_class_for_catalog_type[catalog_type]
+    dask_df, dask_df_pixel_map = _load_dask_df_and_map(hc_catalog, config)
+    catalog = Catalog(dask_df, dask_df_pixel_map, hc_catalog)
+    if config.search_filter is not None:
+        catalog = catalog.search(config.search_filter)
+    if isinstance(config.margin_cache, MarginCatalog):
+        catalog.margin = config.margin_cache
+        if config.search_filter is not None:
+            # pylint: disable=protected-access
+            catalog.margin = catalog.margin.search(config.search_filter)
```
camposandro, line 152: When the margin is provided as an object we're applying the spatial filters but the other parquet arguments (e.g. columns, filters) are unfortunately being ignored. If the margin cache is provided as a path (second branch of this if) the parquet arguments are applied because they are pushed down to the `read_parquet_file_to_pandas` call. Could we just assume that if the margin is provided as an object, it has already been preprocessed?

// This is not related to the refactoring and can definitely be handled in a separate issue!

tests/lsdb/loaders/hats/test_read_hats.py
```diff
@@ -49,6 +48,20 @@ def test_read_hats_no_pandas(small_sky_order1_no_pandas_dir, assert_divisions_ar
...
+        small_sky_xmatch_dir,
+        margin_cache=small_sky_xmatch_margin_dir,
+        columns=["ra", "dec"],
+        filters=[("ra", ">", 300)],
+        engine="pyarrow",
+    )
+    assert isinstance(catalog, lsdb.Catalog)
+    filtered_cat = catalog.compute()
+    assert all(catalog.columns == ["ra", "dec"])
+    assert np.all(filtered_cat["ra"] > 300)
```
camposandro, line 62: Can we also assert that the margin has the right columns and has been filtered?

src/lsdb/loaders/hats/read_hats.py
delucchi-cmu: We had discussed last week to just remove the option to provide the margin as an object, and only as a path. I think it addresses enough of the mismatch concerns, and is also likely the more common route for users.

PR #475, opened 2024-10-29
Remove margin object from `read_hats`

### LINKED ISSUE #473: Remove code path for lsdb.read_hats to provide a margin object
If the call to `lsdb.read_hats(..., margin=margin_object)` is used, it's possible (and easy) for the pre-loaded margin catalog to be inconsistent with the object catalog. It may have different columns, or have different spatial filters applied. 

Initially, we'd been using a pattern of 1) load the margin catalog, and 2) pass the margin catalog to the primary catalog loading, but this is not intuitive for users, and it makes more sense to pass in the path and load both at the same time.

camposandro:
Remove code path that allows to provide a margin object to `read_hats`. Adds a simple validation to make sure that the loaded margin has a compatible schema. Closes #473.

### DIFF
--- a/src/lsdb/loaders/hats/read_hats.py
+++ b/src/lsdb/loaders/hats/read_hats.py
@@ -10,6 +10,7 @@ import numpy as np
 import pyarrow as pa
 from hats.catalog import CatalogType
 from hats.catalog.healpix_dataset.healpix_dataset import HealpixDataset as HCHealpixDataset
+from hats.io import paths
 from hats.io.file_io import file_io
 from hats.pixel_math import HealpixPixel
 from hats.pixel_math.healpix_pixel_function import get_pixel_argsort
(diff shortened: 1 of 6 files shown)


src/lsdb/loaders/hats/read_hats.py
```diff
@@ -146,17 +146,26 @@ def _load_object_catalog(hc_catalog, config):
...
 
+def _validate_margin_catalog(margin_hc_catalog, hc_catalog):
+    """Validate that the margin catalog and the main catalog are compatible"""
+    pixel_columns = [paths.PARTITION_ORDER, paths.PARTITION_DIR, paths.PARTITION_PIXEL]
+    margin_pixel_columns = pixel_columns + ["margin_" + column for column in pixel_columns]
+    catalog_schema = pa.schema([field for field in hc_catalog.schema if field.name not in pixel_columns])
+    margin_schema = pa.schema(
+        [field for field in margin_hc_catalog.schema if field.name not in margin_pixel_columns]
+    )
+    if not catalog_schema.equals(margin_schema):
```
delucchi-cmu, line 165: Can we exercise this in test? e.g. with 

`lsdb.read_hats(small_sky_object, margin=small_sky_source_margin)`

delucchi-cmu (approved):

PR #480, opened 2024-10-31
Add catalog validation tutorial

### LINKED ISSUE #386: Docs: explain Dask concepts and Catalog representation better
~~From `lincc-lsdb` slack channel, Gregory Dubois-Felsmann asked how do we get list of the columns for a catalog and highlighted that docs are not really helpful in discovering `Catalog.columns` attribute~~

It turns, that the real issue Gregory had is in understanding that `read_hipscat()` loads metadata only and what is actually shown when we run `display(catalog)` in Jupyter.

### LINKED ISSUE #392: Docs: add a catalog validation tutorial
This should call the new LSDB validation method (which in turn calls the hipscat validation method).

In addition, provide code snippets to support some strategies that would require more user input:

- comparing data to truth values
- cone search results
- user-supplied min/max check against metadata footers
- per-column aggregations (e.g. sum, mean)

delucchi-cmu:
Closes #392.

Partially resolves #386 (documents the `Catalog.schema` attribute).

### DIFF
--- a/docs/tutorials.rst
+++ b/docs/tutorials.rst
@@ -26,6 +26,7 @@ A more in-depth look into how LSDB works
     :name: Advanced Topics
 
     Topic: Import catalogs <tutorials/import_catalogs>
+    Topic: Manual catalog verification <tutorials/manual_verification>
     Topic: Accessing Remote Data <tutorials/remote_data>
     Topic: Margins <tutorials/margins>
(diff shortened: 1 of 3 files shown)


delucchi-cmu:
See preview at https://lsdb--480.org.readthedocs.build/en/480/tutorials/manual_verification.html

camposandro (approved):

PR #482, opened 2024-10-31
Some grammar and clarity updates

OliviaLynn:
Misc notes:
- Replacing performance testing plot image with a typo in the title
- I'm ambivalent to the actual case of the tutorial titles, would just prefer consistency (either all title case or all only-first-word-capitalized)


delucchi-cmu (approved):
Thank you!

PR #483, opened 2024-10-31
"Getting data into LSDB" should include a reference to the import topic

### LINKED ISSUE #456: Getting data into LSDB is not showing how to get data into LSDB
First tutorial that is `Getting data into LSDB` (https://docs.lsdb.io/en/latest/tutorials/getting_data.html) automatically shows how to import very large catalogs from data.lsdb.io. 

More realistically an user will try to import a mock catalogs or a small test dataset. At minimum there should be a prominent link to `Topics:Import catalogs` page that we show later.

gitosaurus:
Resolves #456

### DIFF
--- a/docs/developer/contributing.rst
+++ b/docs/developer/contributing.rst
@@ -2,7 +2,7 @@ Contributing to LSDB
 ===============================================================================
 
 Installation from Source
----------------------
+------------------------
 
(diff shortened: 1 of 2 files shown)


smcguire-cmu (approved):
Looks good! Can you just run the clear notebooks precommit hook locally and commit those changes so the checks all pass?

PR #488, opened 2024-11-01
Address warning on writing catalog.

### LINKED ISSUE #478: UserWarning: Expected `int` but got `int64` with value `np.int64(2)
Running `ztf_x_gaia.to_hats("./lsdb_out/ztf_x_gaia")` from the[ Getting Started guide](https://docs.lsdb.io/en/latest/getting-started.html#saving-the-result) does write files, but produces the following warning:
```
/sdf/home/o/olynn/.conda/envs/lsdb/lib/python3.11/site-packages/pydantic/main.py:390: UserWarning: Pydantic serializer warnings:
  Expected `int` but got `int64` with value `np.int64(2)` - serialized value may not be as expected
  return self.__pydantic_serializer__.to_python(
```
A very quick and simple visual comparison of the original catalog against the saved-and-read-back-in both do show the same values.

However, still worth either further investigation or catching the warning (or at least acknowledging it in the Getting Started docs) so new users don't think things are broken.

delucchi-cmu:
Closes #478 

I was looking into something unrelated, and observed this and got annoyed.

### DIFF
--- a/src/lsdb/io/to_hats.py
+++ b/src/lsdb/io/to_hats.py
@@ -112,7 +112,7 @@ def to_hats(
         catalog.hc_structure,
         base_catalog_path,
         catalog_name if catalog_name else catalog.hc_structure.catalog_name,
-        total_rows=np.sum(counts),
+        total_rows=int(np.sum(counts)),
     )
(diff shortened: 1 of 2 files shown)


OliviaLynn (approved):
Nice

PR #493, opened 2024-11-06
Update nested-dask to v0.3.0

hombit:
Currently we have a single test failure when querying with column names that include a space. Here is the corresponding Nested-Pandas issue: https://github.com/lincc-frameworks/nested-pandas/issues/174


gitosaurus (approved):
LGTM!

delucchi-cmu (approved):

PR #503, opened 2024-11-18
Add MOC Filter

smcguire-cmu:
Adds the `MOCSearch` class, and the `search_moc` method to the catalog that allows filtering a catalog by a mocpy object.

Also adds a default `search` implementation to the `HealpixDataset` class. Previously searching was only possible on the catalog subclass.

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -194,6 +194,22 @@ class HealpixDataset(Dataset):
         ddf_partition_map = {pixel: i for i, pixel in enumerate(filtered_pixels)}
         return ddf_partition_map, filtered_partitions_ddf
 
+    def search(self, search: AbstractSearch):
+        """Find rows by reusable search algorithm.
+
+        Filters partitions in the catalog to those that match some rough criteria.
+        Filters to points that match some finer criteria.
+
+        Args:
+            search (AbstractSearch): Instance of AbstractSearch.
+
+        Returns:
+            A new Catalog containing the points filtered to those matching the search parameters.
+        """
+        filtered_hc_structure = search.filter_hc_catalog(self.hc_structure)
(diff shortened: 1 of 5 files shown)


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -194,6 +194,22 @@ def _perform_search(
         ddf_partition_map = {pixel: i for i, pixel in enumerate(filtered_pixels)}
         return ddf_partition_map, filtered_partitions_ddf
 
+    def search(self, search: AbstractSearch):
```
camposandro, line 197: Can we delete the `search` method from the Catalog object entirely and just use this one to reduce duplication? We'd just need to be careful about margin filtering.

camposandro: We could also remove it from `MarginCatalog`:

https://github.com/astronomy-commons/lsdb/blob/34f7d9d518533f52e25f1d21d5db02eec26fefde/src/lsdb/catalog/margin_catalog.py#L28

src/lsdb/core/search/moc_search.py
```diff
@@ -0,0 +1,34 @@
...
+
+if TYPE_CHECKING:
+    from lsdb.types import HCCatalogTypeVar
+
+
+class MOCSearch(AbstractSearch):
+    """Filter the catalog by HEALPix pixels.
+
+    Filters partitions in the catalog to those that are in a specified pixel set.
+    Does not filter points inside those partitions.
```
camposandro, line 20: Filter the catalog by a MOC

src/lsdb/catalog/dataset/healpix_dataset.py
smcguire-cmu: The catalog has the margin logic, but it should call the superclass, but yeah removing the margin is a good idea

camposandro (approved):

PR #508, opened 2024-11-20
Fix empty margin catalogs in `from_dataframe`

### LINKED ISSUE #419: Return empty margin catalog instead of None with from_dataframe
**Bug report**

`from_dataframe()` should assign empty margin catalog instead of `None` when margin is empty. Originated from this discussion:
https://github.com/astronomy-commons/lsdb/issues/416#issuecomment-2359150932

It looks like it is the similar to #329

**Before submitting**
Please check the following:

- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.

camposandro:
Fixes issue with empty catalogs in `from_dataframe` being assigned to None. It also fixes the margin catalog schema, which was incorrectly assigned to the schema of the main catalog (see `_create_margin_schema`). Closes #419.


src/lsdb/catalog/margin_catalog.py
```diff
@@ -23,3 +25,20 @@ def __init__(
...
     ):
         super().__init__(ddf, ddf_pixel_map, hc_structure)
+
+
+def _validate_margin_catalog(margin_hc_catalog, hc_catalog):
+    """Validate that the margin and main catalogs have compatible schemas. The order of
+    the pyarrow fields should not matter."""
+    expected_margin_schema = _create_margin_schema(hc_catalog.schema)
+    # Compare the field set for the schemas. They should match.
+    if set(margin_hc_catalog.schema) != set(expected_margin_schema):
```
smcguire-cmu, line 35: I don't like this set comparison because duplicated columns won't be detected. Either do a list and sort them then compare, or check the length of the sets are equal to the length of the schema

src/lsdb/loaders/dataframe/margin_catalog_generator.py
```diff
@@ -71,23 +71,34 @@ def _resolve_margin_order(self):
...
 
-    def create_catalog(self) -> MarginCatalog | None:
+    def create_catalog(self) -> MarginCatalog:
         """Create a margin catalog for another pre-computed catalog
 
         Returns:
             Margin catalog object, or None if the margin is empty.
         """
+        if self.margin_order < 0 and self.margin_threshold <= 0:
+            return self.create_empty_margin_catalog()
```
smcguire-cmu, line 81: In this case we're actually not generating a margin catalog, so this should return None

src/lsdb/catalog/margin_catalog.py
smcguire-cmu: Actually, the length thing won't work because we want to allow duplicates if both schemas have them

src/lsdb/loaders/dataframe/from_dataframe.py
```diff
@@ -20,7 +20,7 @@ def from_dataframe(
     partition_size: int | None = None,
     threshold: int | None = None,
     margin_order: int | None = -1,
-    margin_threshold: float | None = 5.0,
+    margin_threshold: float = 5.0,
```
smcguire-cmu, line 23: I think this should still be allowed to be None to generate no margin.

smcguire-cmu (approved):

PR #511, opened 2024-11-25
Implement map type and merging

### LINKED ISSUE #486: Implement MAP catalog type
See: https://github.com/astronomy-commons/hats/issues/412

### LINKED ISSUE #487: Implement map merging
See https://github.com/astronomy-commons/lsdb/issues/486

Implement mechanism to join a point-source catalog (e.g. object catalog) to a continuous map type.

Implement as a new operation, `merge_map` (name is negotiable) that, similar to a cross-match, allows for differing algorithms to suit the very different needs of map varieties. e.g.

- Photo-D style priors: would also want the custom function to perform some map_partitions style operations to reduce the data before leaving the routine.
- Pixel Mask: effectively filters out values in the point-source catalog
- Extinction map: adds a single (or a couple, e.g. value and uncertainty) floating point values per row in point-source catalog.

delucchi-cmu:
Closes #486. Closes #487.


src/lsdb/dask/merge_map_catalog_data.py
```diff
@@ -0,0 +1,124 @@
...
+)
+from lsdb.types import DaskDFPixelMap
+
+if TYPE_CHECKING:
+    from lsdb.catalog import Catalog, MapCatalog
+
+
+# pylint: disable=too-many-arguments, unused-argument
+@dask.delayed
+def perform_merge_map(
```
gitosaurus, line 28: Given that this performs a `merge_asof`, why not call it `perform_merge_maps_asof` or similar?  Is `asof` pretty obviously implied through this family of functions?  (Seems like it may be.)

gitosaurus (approved):
LGTM

src/lsdb/dask/merge_map_catalog_data.py
delucchi-cmu: This doesn't perform a `merge_asof`, but I had copy/pasted the `merge_asof` as a starting point, and never updated the docstrings!

This functionality is more akin to a custom cross-match, in that some custom function will be applied to spatially matching pairs of partitions.

PR #513, opened 2024-11-25
Update margin docs notebook to use new plotting functions

smcguire-cmu:
Uses plot_pixels and plot_points instead of custom plotting function in margin tutorial.


camposandro (approved):
LGTM! Just wondering if it is possible to add a label for each set of points? When calling `plot_points(c="green", marker="+", label="...")`.

smcguire-cmu:
Good idea! I added a legend with those to the plots

PR #514, opened 2024-11-25
Update Catalog Lazy Representation

smcguire-cmu:
- Changes index column from `healpix_29` divisions to healpix pixels
- Adds warning that catalog is only lazily read


camposandro (commented):
Awesome, looks great! Can we just update the images of the [Getting Started](https://lsdb--514.org.readthedocs.build/en/514/getting-started.html) docs to show the current representation? the second image is unrelated to this change but it still shows a `_hipscat_index` column.

camposandro (approved):

Issue #529, opened 2024-12-14
Add metadata to Gaia and ZTF catalogs


troyraen:
**Feature request**

We are developing a prototype web service comprised of an LSDB query server backend and a Firefly frontend. For the purposes of demonstration, it would be nice to have the following metadata:

- Column descriptions
- Column units
- Specification of "default" columns (used to determine which columns are selected by default)

added to the following catalogs found at https://data.lsdb.io/:

- [x] Gaia DR3 (GAIA_SOURCE)
- [x] ZTF DR14 (objects)
- [x] ZTF DR14 (sources) 

We will use the metadata to populate the respective columns in the bottom panel of the screenshot below.

<img width="1403" alt="Screenshot 2024-12-13 at 6 17 36 PM" src="https://github.com/user-attachments/assets/1b8d80cf-57f6-4217-9b32-a4d919e6a13a" />

The `default_columns` property of [TableProperties](https://hats.readthedocs.io/en/stable/autoapi/hats/catalog/dataset/table_properties/) looks like the intended place to specify those. I don't see an obvious place for column descriptions and units at the moment. Ultimately, we'd like the metadata in the VOTable-in-Parquet format that is currently under development. In the meantime, we ought to be able to pull it from wherever you think is easiest/best to put it. We're using the hats library to get at the metadata, though we're not limited to that.

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [x] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [x] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

delucchi-cmu:
Can you please list the expected default columns for the three catalogs specified?

troyraen:
There's no specific expectation but here's my suggestions.

Gaia
- solution_id
- designation
- source_id
- ref_epoch
- ra
- ra_error
- dec
- dec_error
- parallax
- parallax_error
- pm
- pmra
- pmra_error
- pmdec
- pmdec_error
- phot_g_n_obs
- phot_g_mean_flux
- phot_g_mean_flux_error
- phot_g_mean_mab
- phot_bp_n_obs
- phot_bp_mean_flux
- phot_bp_mean_flux_error
- phot_bp_mean_mag
- phot_rp_n_obs
[...]

PR #543, opened 2025-01-22
Use default_columns property from hats catalog if available in read_hats

### LINKED ISSUE #267: Read catalog - get only positional coordinates by default
**Feature request**

When reading a catalog with `read_hipscat` change the default to only read in `ra, dec` coordinates, i.e., spatial coordinates, and require user to specify further columns. This is to avoid extensive loading when exploring the datasets (e.g., loading 10, 20, 50 columns of a catalog when only looking at the number of objects or similar). 

**Before submitting**
Please check the following:

smcguire-cmu:
Uses the property added in response to #529 when loading a hats catalog

Closes #267


tests/data/small_sky_order1_default_columns/properties
```diff
@@ -0,0 +1,15 @@
...
+hats_col_dec=dec
+hats_max_rows=1000000
+hats_order=1
+moc_sky_fraction=0.08333
+hats_builder=hats-import v0.3.6.dev26+g40366b4
+hats_creation_date=2024-10-15T14\:47UTC
+hats_estsize=39
+hats_release_date=2024-09-18
+hats_version=v0.1
+hats_cols_default=ra dec id ra_error dec_error
```
delucchi-cmu, line 15:
```suggestion
hats_cols_default=ra dec id 
```

Otherwise, the default columns are just all of the columns and that's not very interesting.

tests/data/small_sky_order1_default_columns/dataset/_metadata
delucchi-cmu, line 1: Since everything inside `dataset/` is identical to `/data/small_sky_order1`, could we just provide a symlink here instead of data duplication?

And include the data generation in the `generate_data` notebook?

tests/data/small_sky_order1_default_columns/properties
smcguire-cmu: Oh sure! I was thinking that excluding the 'Norder', 'Pix', and 'Dir' columns would be enough, but I wasn't sure if they 'counted' as columns that would normally be loaded for this.

delucchi-cmu (approved):

PR #568, opened 2025-02-06
Pass args and kwargs to derive meta on `map_partitions`

### LINKED ISSUE #564: map_partitions should pass args and kwargs when deriving meta
**Bug report**

`.map_partitions(func, func_argument="value")` fails if `func` requires `func_argument`:

```python
import lsdb

def my_func(df, a):
    return df.assign(b=df["ML_CTS_P6"] * a)

catalog = lsdb.read_hats('https://data.lsdb.io/hats/erosita')
catalog.map_partitions(my_func, 123)
```

```
TypeError                                 Traceback (most recent call last)
Cell In[6], line 7
      4     return df.assign(b=df["ML_CTS_P6"] * a)
      6 catalog = lsdb.read_hats('https://data.lsdb.io/hats/erosita')
----> 7 catalog.map_partitions(my_func, 123)

File ~/projects/lincc-frameworks/lsdb/src/lsdb/catalog/dataset/healpix_dataset.py:280, in HealpixDataset.map_partitions(self, func, meta, include_pixel, *args, **kwargs)
    278     meta = func(self._ddf._meta.copy(), HealpixPixel(0, 0))
    279 else:
--> 280     meta = func(self._ddf._meta.copy())
    281 if meta is None:
    282     raise ValueError(
    283         "func returned None for empty DataFrame input. The function must return a value, changing"
    284         " the partitions in place will not work. If the function does not work for empty inputs, "
    285         "please specify a `meta` argument."
    286     )

TypeError: my_func() missing 1 required positional argument: 'a'
```

The problem is that here we don't pass `*args, **kwargs` when calling a function to produce a meta
https://github.com/astronomy-commons/lsdb/blob/96b41f91fb152de7a034535081ccbcffb8fdc47d/src/lsdb/catalog/dataset/healpix_dataset.py#L278-L280

**Before submitting**
Please check the following:

camposandro:
Add the missing args and kwargs for the user-defined function when deriving the meta for `map_partitions`. Closes #564.

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -275,9 +275,9 @@ class HealpixDataset(Dataset):
         """
         if meta is None:
             if include_pixel:
-                meta = func(self._ddf._meta.copy(), HealpixPixel(0, 0))
+                meta = func(self._ddf._meta.copy(), HealpixPixel(0, 0), *args, **kwargs)
             else:
-                meta = func(self._ddf._meta.copy())
+                meta = func(self._ddf._meta.copy(), *args, **kwargs)
             if meta is None:
                 raise ValueError(
                     "func returned None for empty DataFrame input. The function must return a value, changing"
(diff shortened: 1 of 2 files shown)


hombit (approved):
Thank you

PR #574, opened 2025-02-11
Update performance tutorial with updated analysis

smcguire-cmu:


hombit (changes requested):
1. Could you please explain purple curve (RA and Dec columns)
2. Could you please move your code to https://github.com/lincc-frameworks/notebooks_lf?

hombit (approved):

PR #596, opened 2025-03-01
Operations update metadata correctly

### LINKED ISSUE #143: Refactor common catalog creation logic
There's some very similar lines of code when we generate a new catalog after doing a search, crossmatch, or join, to generate the divisions, make the dask df, then make the catalog.

```
    divisions = get_pixels_divisions(list(partition_map.keys()))
    ddf = dd.from_delayed(partitions, meta=meta_df, divisions=divisions)
    ddf = cast(dd.DataFrame, ddf)
```

In the crossmatch and join methods where most of the logic live outside of the catalog class, a catalog can't be created due to circular imports. There could be some other way to reuse this code between them like moving some of the merging logic inside the catalog class, or moving the search logic outside.

### LINKED ISSUE #178: Inherited catalog methods should use `self.__class__()` to init returned objects

### LINKED ISSUE #566: `to_hats` writes old default columns to properties, even if column names have changed
We should probably set the default columns property to None as soon as anything is done to the catalog.

smcguire-cmu:
Refactors catalogs to create returned objects using a new `_create_updated_dataset` that sets `total_rows` to 0 and updates the `schema` and `default_columns` properties in the `hc_structure`. 

Fixes #566 fixes #178 fixes #143

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -1,3 +1,4 @@
+import hats as hc
 from __future__ import annotations
 
 import warnings
@@ -129,14 +130,26 @@ class HealpixDataset(Dataset):
         hc_structure: HCHealpixDataset | None = None,
         updated_catalog_info_params: dict | None = None,
     ) -> Self:
-        """Creates an"""
+        """Creates a new copy of the catalog, updating any provided arguments
+
+        Shallow copies the ddf and ddf_pixel_map if not provided. Creates a new hc_structure if not provided.
+        Updates the hc_structure with any provided catalog info parameters, resets the total rows, removes
+        any default columns that don't exist, and updates the pyarrow schema to reflect the new ddf.
+
+        Args:
+            ddf (nd.NestedFrame): The catalog ddf to update in the new catalog
+            ddf_pixel_map (DaskDFPixelMap): The partition to healpix pixel map to update in the new catalog
+            hc_structure (hats.HealpixDataset): The hats HealpixDataset object to update in the new catalog
+            updated_catalog_info_params (dict): The dictionary of updates to the parameters of the hats
+                dataset object's catalog_info
+        Returns:
+            A new dataset object with the arguments updated to those provided to the function, and the
+            hc_structure metadata updated to match the new ddf
+        """
         ddf = ddf or self._ddf
-        if ddf_pixel_map is None:
-            ddf_pixel_map = self._ddf_pixel_map
-        if hc_structure is None:
-            hc_structure = self.hc_structure
-        if updated_catalog_info_params is None:
-            updated_catalog_info_params = {}
+        ddf_pixel_map = ddf_pixel_map or self._ddf_pixel_map
+        hc_structure = hc_structure or self.hc_structure
+        updated_catalog_info_params = updated_catalog_info_params or {}
         if (
             "default_columns" not in updated_catalog_info_params
             and hc_structure.catalog_info.default_columns is not None
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -130,8 +130,7 @@ class HealpixDataset(Dataset):
         updated_catalog_info_params: dict | None = None,
     ) -> Self:
         """Creates an"""
-        if ddf is None:
-            ddf = self._ddf
+        ddf = ddf or self._ddf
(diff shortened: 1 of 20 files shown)


delucchi-cmu:
This makes progress on https://github.com/astronomy-commons/lsdb/issues/178. Does it close it?

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -104,20 +104,55 @@ def _repr_divisions(self):
...
         )
 
+    def _create_updated_dataset(
+        self,
+        ddf: nd.NestedFrame | None = None,
+        ddf_pixel_map: DaskDFPixelMap | None = None,
+        hc_structure: HCHealpixDataset | None = None,
+        updated_catalog_info_params: dict | None = None,
+    ) -> Self:
+        """Creates an"""
```
delucchi-cmu, line 132: ??

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -104,20 +104,55 @@ def _repr_divisions(self):
...
+        if ddf is None:
+            ddf = self._ddf
```
delucchi-cmu, line 134:
```suggestion
        ddf = ddf or self._ddf
```

And in a half-dozen places following.

Maybe even update the docstring to say that this method exists to coalesce user-provided arguments with object members.

src/lsdb/io/to_hats.py
```diff
@@ -87,6 +88,9 @@ def to_hats(
         catalog (HealpixDataset): A catalog to export
         base_catalog_path (str): Location where catalog is saved to
         catalog_name (str): The name of the output catalog
+        default_columns (list[str]): A metadata property with the list of the columns in the catalog to be
+            loaded by default. By default, uses the default columns from the original hats catalogs if they
```
delucchi-cmu, line 92:
```suggestion
            loaded by default. Uses the default columns from the original hats catalogs if they
```

delucchi-cmu:
> This makes progress on #178. Does it close it?

Maybe also most of https://github.com/astronomy-commons/lsdb/issues/143?

tests/lsdb/loaders/hats/test_read_hats.py
```diff
@@ -34,23 +23,26 @@ def test_read_hats(small_sky_order1_dir, small_sky_order1_hats_catalog, assert_d
...
     assert isinstance(catalog.compute(), npd.NestedFrame)
-    assert_divisions_are_correct(catalog)
-    assert_index_correct(catalog)
+    helpers.assert_divisions_are_correct(catalog)
+    helpers.assert_index_correct(catalog)
+    helpers.assert_schema_correct(catalog)
 
 
-def test_read_hats_default_cols(small_sky_order1_default_cols_dir, assert_divisions_are_correct):
+def test_read_hats_default_cols(small_sky_order1_default_cols_dir, helpers):
```
delucchi-cmu, line 31: I'd like to see a test that performs some round-trip operations, as mentioned in comment: https://github.com/astronomy-commons/lsdb/issues/566#issuecomment-2683212732

src/lsdb/catalog/dataset/healpix_dataset.py
smcguire-cmu: My greatest docstring. Oops

tests/lsdb/loaders/hats/test_read_hats.py
smcguire-cmu: We have one of those actually! `test_save_catalog_default_columns` and `test_save_catalog_empty_default_columns` do that. We actually had round trip tests before, but they weren't using catalogs with default columns. But I also just added `test_save_catalog_default_columns_cross_matched` to test a round trip with a crossmatched catalog too.

src/lsdb/catalog/dataset/healpix_dataset.py
smcguire-cmu: Oh actually it seems like a nonempty ddf raises an error rather than being true. I've replaced this with an inline if.

smcguire-cmu:
I think it does close #178! There's maybe a bit more refactoring that could be done with #143, but I think it's mostly there and enough to close it. But I can leave it open if you think it's better to.

delucchi-cmu (approved):

PR #604, opened 2025-03-05
Re-generate test data, and update expectations.

delucchi-cmu:
See similar change (and justifications) on https://github.com/astronomy-commons/hats/pull/467

This also tidies up the imports on the generate_data notebook!!


tests/data/small_sky_source_margin/properties
```diff
@@ -1,15 +1,15 @@
 #HATS catalog
 obs_collection=small_sky_source_margin
 dataproduct_type=margin
-hats_nrows=395
+hats_nrows=1717
```
gitosaurus, line 4: Large increase in the number of rows, here and throughout. Increased size of small sky?

tests/data/square_map/properties
```diff
@@ -5,8 +5,8 @@ hats_nrows=12
 hats_max_rows=1000000
 hats_order=0
 moc_sky_fraction=1.00000
-hats_builder=hats-import v0.4.2.dev1+g70355fa
-hats_creation_date=2024-11-25T15\:17UTC
-hats_estsize=98350
+hats_builder=hats-import v0.4.6.dev1+gf00cd7a
+hats_creation_date=2025-03-05T19\:01UTC
+hats_estsize=34
```
gitosaurus, line 10: Conversely, this seems like a big drop in size.

gitosaurus (approved):
LGTM! Just curious about a couple of what seem like large changes in size.

tests/data/small_sky_source_margin/properties
delucchi-cmu: In the case of margins, it's mostly due to sweeping changes to margin generation made in November that removes the fine filtering in generation, and instead only performs a coarse healpix tile based filtering.

PR #619, opened 2025-03-12
Add wrapper for `sort_values`

### LINKED ISSUE #601: Add Catalog.sort_values
**Feature request**

`sort_values` was implemented in both nested-pandas and nested-dask. I think we should support it here, at least for nested columns, so it is easy to sort time-series.

**Before submitting**
Please check the following:

- [ ] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.

camposandro:
Add a wrapper for `NestedFrame.sort_values` to enable catalog sorting by base and nested columns. Resolves #601.


hombit:
I'm not sure if we want to allow "base-column" sorting here: user may accidentally shuffle the whole Dask data frame

hombit:
Alternatively we may allow base columns, but sort with map_partitions

camposandro:
Thanks Kostya, that is a good point! I feel more inclined to either issue a warning about shuffling when sorting on base column, or remove that option entirely. Sorting per partition could potentially be confusing?

src/lsdb/catalog/catalog.py
```diff
@@ -856,3 +857,92 @@ def reduce(self, func, *args, meta=None, **kwargs) -> Catalog:
...
+            na_position=na_position,
+            partition_size=partition_size,
+            sort_function=sort_function,
+            sort_function_kwargs=sort_function_kwargs,
+            upsample=upsample,
+            ignore_index=ignore_index,
+            shuffle_method=shuffle_method,
+            **options,
+        )
+        if self.margin is not None:
```
delucchi-cmu, line 934: Is the margin still meaningful if the full catalog has been sorted by something other than healpix?

camposandro: Thanks Melissa, you're right, in that case it's not meaningful! This takes us back to the question, should we allow for base column sorting? I think our current options are:

1. Allow for sorting of base columns with shuffling (invalidates margins);
2. Allow for sorting of base columns via `map_partitions` (keeps margins);
3. Only allow for sorting of nested columns (also keeps margins).

The first two do not feel right/intuitive:

1. It tampers with the healpix partitioning. One is not able to use the resulting catalog for spatial operations (e.g. search filters and crossmatching).
2. Users would expect to have catalog-level sorting but this would only provide partition-level sorting?

I'm leaning toward option 3., potentially naming the method to `sort_nested_values`, for added clarity?

Tagging @hombit @dougbrn, we'd love to hear your opinion!

hombit: I vote for `sort_nested_values`. I think we should keep margins and sort them in the same way

dougbrn: I agree with your leaning @camposandro, 1. seems directly against the design of lsdb/hats, 2. would have a potential advantage of allowing second-order sorting within a partition, but I think you could add a more intentional utility for this down the road that does it at the row-group level. 3. feels right,  and a rename would be a good idea there, `sort_nested_values` would be my first thought as well.

dougbrn (approved):
Looks good to me!

src/lsdb/catalog/catalog.py
```diff
@@ -856,3 +856,50 @@ def reduce(self, func, *args, meta=None, **kwargs) -> Catalog:
...
+    def sort_nested_values(
+        self,
+        by: str | list[str],
+        ascending: bool | list[bool] = True,
+        na_position: Literal["first"] | Literal["last"] = "last",
+        ignore_index: bool | None = False,
+        **options,
+    ) -> Catalog:
+        # pylint: disable=duplicate-code
+        """Sort nested columns for each row in the catalog.
```
delucchi-cmu, line 869:
```suggestion
        """Sort nested columns for each row in the catalog.
        
        Note that this does NOT sort rows, only nested values within rows.
```

delucchi-cmu: This might be overkill, but it would help me to see this kind of note.

src/lsdb/catalog/catalog.py
```diff
@@ -856,3 +856,50 @@ def reduce(self, func, *args, meta=None, **kwargs) -> Catalog:
...
+
+        Args:
+            by: str or list[str]
+                Column(s) to sort by.
+            ascending: bool or list[bool], optional
+                Sort ascending vs. descending. Defaults to True. Specify list for
+                multiple sort orders. If this is a list of bools, must match the
+                length of the by.
```
delucchi-cmu, line 877:
```suggestion
                length of the `by`.
```

tests/data/generate_data.ipynb
delucchi-cmu, line 1: Thanks for adding a nested catalog! We should (at a later time) add more tests against this data!

PR #623, opened 2025-03-14
Parametrize test instead of for loop.

delucchi-cmu:
This method was tricky to debug in something else I'm working on. Made much easier by not having tests inside a for-loop.


dougbrn (approved):
Looks good!

PR #625, opened 2025-03-18
Add an intro to the use of map_partitions.

### LINKED ISSUE #536: Example of using map_partition in the documentation
Small example showing how one would use map_partitions

gitosaurus:
Resolves #536.


docs/tutorials/pre_executed/map_partitions.ipynb
```diff
@@ -0,0 +1,7765 @@
+{
```
smcguire-cmu, line 1: <p>Can you add an explanation of the <code>include_pixel</code> parameter, and how that will introduce the healpix pixel as the second parameter.</p><p> </p><p>Also can you specify that it's one row per partition if you do a reduction that returns one row, since you could return more than one row per partition. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/astronomy-commons/lsdb/pull/625/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

smcguire-cmu, line 1: <p>Can you hide these info logging messages? I think this at the start of the notebook will do it</p>

<pre class="ql-syntax" spellcheck="false">import logging
logging.getLogger().setLevel(logging.CRITICAL)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/astronomy-commons/lsdb/pull/625/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

smcguire-cmu, line 1: <p>I think this is a bit more of a common use case for map_partitions than the reductions, so it might be worth putting this before the generating stats part above.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/astronomy-commons/lsdb/pull/625/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

gitosaurus: <p>That certainly should have done the trick! But it didn't, not even with restarting the kernel. INFO messages continue to pour out from <code>distributed.scheduler</code>. This is on data.lsst.cloud . I will look around and see if there's some other magic spell; we can't be the only people who want this.</p>

gitosaurus: <p>Amazingly, it's this much trouble to finally get it to quiet down:</p><p> </p>

<pre class="ql-syntax" spellcheck="false">import dask
dask.config.set({"logging.distributed": "error"})

# This also has to be done, for the above to be effective
logger = logging.getLogger("distributed")
logger.setLevel(logging.ERROR)

import warnings
# Finally, suppress the specific warning about Dask dashboard port usage
warnings.filterwarnings("ignore", message="Port 8787 is already in use.")
</pre>

gitosaurus: <p>Some of this is tucked into comments, but I agree, it all needs to be clearer and up in the markdown cells that introduce it. Done!</p>

gitosaurus: <p>Good idea. Reorganized.</p>

smcguire-cmu (approved):

PR #628, opened 2025-03-18
Don't assume Norder/Dir/Npix columns

delucchi-cmu:
This will also suppress these columns when catalogs are created via `from_dataframe`. Regenerates test data to match expectations.

See https://github.com/astronomy-commons/hats/issues/367


camposandro (approved):
Sorry for delay in the review!

PR #629, opened 2025-03-18
Add lsdb.crossmatch(left, right, ....) method, which supports frames and catalogs

### LINKED ISSUE #618: Dataframe/catalog crossmatching when catalog has margin
PR #602 enabled crossmatching a pandas dataframe with an lsdb catalog, eg:

```python
my_catalog_x_dataframe = my_catalog.crossmatch_dataframe(my_dataframe, ...)
```

However, this **only supports crossmatching where the dataframe is on the right**.

This matters for two main reasons:
1. margin caches (specified for the right-most item)
2. one-to-many/many-to-one relationships ("with a small custom catalog user usually wants to get a single match per each of their catalog object.")

How would we want to let the users **specify a crossmatch where the dataframe is on the left**?

**1. Only allow when using `Catalog.crossmatch_dataframe(my_dataframe, my_catalog_with_margin, ...)` syntax**
- Makes users explicitly list out both members of the crossmatch in the order that we're used to seeing (where the right-most has the margin)
- Potentially confusing, as the `Catalog.crossmatch(a, b, ...) syntax is rarely used elsewhere in code/docs

**2.** (My other suggested options are removed now that I have a little more context on this)

**3. ???**
- Open to suggestions!

OliviaLynn:
Closes #618

Pre-commit check failure is due to:
```
pylint (python files in src/)............................................Failed
- hook id: pylint
- duration: 28.39s
- exit code: 8

************* Module lsdb.loaders.dataframe.from_dataframe
src/lsdb/loaders/dataframe/from_dataframe.py:1:0: R0801: Similar lines in 2 files
==lsdb.catalog.dataset.healpix_dataset:[845:856]
==lsdb.core.plotting.plot_points:[68:79]
        fig, ax, wcs = initialize_wcs_axes(
            projection=projection,
            fov=fov,
            center=center,
            wcs=wcs,
            frame_class=frame_class,
            ax=ax,
            fig=fig,
            figsize=(9, 5),
        )
 (duplicate-code)
```
Unsure if this will block merging. 

If so, would it be alright for me to add `# pylint: disable=duplicate-code` to the files in question and open an issue to investigate/fix duplicate code?


src/lsdb/core/crossmatch/crossmatch.py
```diff
@@ -0,0 +1,97 @@
...
+from lsdb.core.crossmatch.abstract_crossmatch_algorithm import AbstractCrossmatchAlgorithm
+from lsdb.core.crossmatch.crossmatch_algorithms import BuiltInCrossmatchAlgorithm
+from lsdb.loaders.dataframe.from_dataframe import from_dataframe
+
+
+def _validate_and_convert_to_catalog(
+    data: npd.NestedFrame | pd.DataFrame, data_args: dict, suffix: str | None, default_suffix: str
+) -> Catalog:
+    """Validate arguments and convert a DataFrame or NestedFrame to a Catalog."""
+    if not isinstance(data, Catalog):
```
delucchi-cmu, line 16:
```suggestion
    if isinstance(data, Catalog):
        return data
```

Lets you use one fewer level of indentation for the remainder, and makes it clear that no additional logic is performed if it's already an instance of `Catalog`.

src/lsdb/core/crossmatch/crossmatch.py
```diff
@@ -0,0 +1,97 @@
...
+        if not isinstance(data, (pd.DataFrame, npd.NestedFrame)):
+            raise TypeError(f"Argument must be a DataFrame, NestedFrame, or Catalog, not {type(data)}.")
+
+        # Check for "ra" and "dec" columns. If they do not exist, try "RA" and "DEC" before raising an error.
+        if "ra_column" not in data_args:
```
delucchi-cmu, line 21: What if someone just provided `ra_column` as a kwarg to the cross match method?

e.g. 

```
# could use coord_ra on the right frame
crossmatch(real_catalog, some_data, ra_column="coord_ra")
```
or

```
# could use coord_ra on both right and left frame.
crossmatch(some_other_data, some_data, ra_column="coord_ra")
```

Or is that complicating the usage too much?

src/lsdb/core/crossmatch/crossmatch.py
OliviaLynn: Delightful! Thank you

delucchi-cmu (approved):
Suggestion on clarifying behavior of arguments. Otherwise looks good!

PR #632, opened 2025-03-19
Update pyproject.toml

delucchi-cmu:
Update pins for hats and nested versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
     "deprecated",
     "hats>=0.5.0",
     "nested-dask>=0.3.4",
-    "nested-pandas",
+    "nested-pandas>=0.3.8",
     "pyarrow",
     "scipy", # kdtree
 ]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -22,8 +22,8 @@ dependencies = [
     # dask diagnostics is required to spin up the dashboard for profiling.
     "dask[complete]",
     "deprecated",
-    "hats>=0.4.6",
-    "nested-dask>=0.3.0",
+    "hats>=0.5.0",
+    "nested-dask>=0.3.4",
     "nested-pandas",
```
camposandro, line 27: Should we temporarily ensure that nested-pandas is `>=0.3.8`? That will be the default for `nested-dask` in their next release (https://github.com/lincc-frameworks/nested-dask/pull/91), but currently the constraint on their side is quite outdated (v0.3.1) and a lot of bugs have been fixed since then.

PR #635, opened 2025-03-20
Wrap methods for getting catalog-level parquet statistics.

### LINKED ISSUE #550: Enable inspection of catalog per leaf
Enable an user to see distribution of data per leaf. Something like summary statistics that would show:

Healpix id - median ra (or edges of ra distribution) - median dec (or edges of dec distribution) - Number of rows - min, max values of for column values.

delucchi-cmu:
Closes #550


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -182,6 +182,48 @@ def get_ordered_healpix_pixels(self) -> list[HealpixPixel]:
...
+        )
+
+    def per_pixel_statistics(
+        self,
+        use_default_columns: bool = True,
+        exclude_hats_columns: bool = True,
+        exclude_columns: list[str] | None = None,
+        include_columns: list[str] | None = None,
+        include_stats: list[str] | None = None,
+        multiindex=False,
```
gitosaurus, line 211:
```suggestion
        multi_index=False,
```
Minor suggestion.  If this is already the way this term is spelled elsewhere, feel free to ignore.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -788,6 +830,7 @@ def reduce_part(df):
             hc_updates = {"ra_column": "", "dec_column": ""}
         return self._create_updated_dataset(ddf=ndf, updated_catalog_info_params=hc_updates)
 
+    # pylint: disable=duplicate-code
```
gitosaurus, line 833: 😃

gitosaurus (approved):
LGTM!

src/lsdb/catalog/dataset/healpix_dataset.py
delucchi-cmu: Changed in HATS in https://github.com/astronomy-commons/hats/pull/479

PR #659, opened 2025-04-02
New catalog preview methods tail, sample, and random_sample

### LINKED ISSUE #549: Better support for viewing small amount of data beyond head() call
**Feature request**
Better support for viewing small part of data beyond head() call

I suggest:
`tail()`
`sample(n=)` : sample number of rows from a single partition
`random_sample(n=)` : sample number of rows randomly from the whole datasets

gitosaurus:
Add a few more preview methods to `lsdb.Catalog`, allowing a user to get a quick sense of what data is in the catalog.  New methods are:

```python
def tail(self, n: int = 5) -> npd.NestedFrame:
    """Returns a few rows of data from the end of the catalog for previewing purposes.
    """

def sample(self, partition_id: int, n: int = 5) -> npd.NestedFrame:
    """Returns a few randomly sampled rows from a given partition.
    """

def random_sample(self, n: int = 5) -> npd.NestedFrame:
    """Returns a few randomly sampled rows, like self.sample(),
       except that it randomly samples all partitions in order to
       fulfill the rows.
    """
```

Closes #549.

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -287,6 +287,10 @@ class HealpixDataset(Dataset):
         partition = self._ddf.partitions[partition_id]
         # Get the count of rows in the partition.
         pixel_rows = len(partition)
+        if pixel_rows == 0:
+            fraction = 0.0
+            logging.debug("Zero rows in partition %d, returning empty", partition_id)
+            return partition._meta
         fraction = n / pixel_rows
         logging.debug("Getting %d / %d = %f", n, pixel_rows, fraction)
         return partition.sample(frac=fraction).compute()
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -296,7 +296,16 @@ class HealpixDataset(Dataset):
             A pandas DataFrame with up to `n` rows of data.
         """
         dfs = []
-        row_counts = np.array(dask.compute(*[dp.shape[0].to_delayed() for dp in self._ddf.partitions]))
+        if self.hc_structure.catalog_info.total_rows > 0:
+            stats = self.hc_structure.per_pixel_statistics()
+            # These stats are one *row* per pixel.  The number of
+            # columns is permuted, with names like "colname:
+            # row_count".  We only need one representative column.
+            # Assume the first column is satisfactory.
+            rep_col = self.columns[0]
+            row_counts = stats[f"{rep_col}: row_count"].map(int)
+        else:
+            row_counts = np.array(dask.compute(*[dp.shape[0].to_delayed() for dp in self._ddf.partitions]))
         npartitions = len(row_counts)
         rows_per_partition = np.random.multinomial(n, row_counts / row_counts.sum())
         # With this breakdown, we randomly sample rows from each partition
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -227,7 +227,7 @@ class HealpixDataset(Dataset):
             n (int): The number of desired rows.
 
         Returns:
-            A pandas DataFrame with up to `n` rows of data.
+            A NestedFrame with up to `n` rows of data.
         """
         dfs = []
         remaining_rows = n
@@ -249,7 +249,7 @@ class HealpixDataset(Dataset):
             n (int): The number of desired rows.
 
         Returns:
-            A pandas DataFrame with up to `n` rows of data.
+            A NestedFrame with up to `n` rows of data.
         """
         dfs = []
         remaining_rows = n
@@ -277,7 +277,7 @@ class HealpixDataset(Dataset):
         as in unit tests.
 
         Returns:
-            A pandas DataFrame with up to `n` rows of data.
+            A NestedFrame with up to `n` rows of data.
         """
         random.seed(seed)
         # Get the number of partitions so that we can range-check the input argument
@@ -309,7 +309,7 @@ class HealpixDataset(Dataset):
         as in unit tests.
 
         Returns:
-            A pandas DataFrame with up to `n` rows of data.
+            A NestedFrame with up to `n` rows of data.
         """
         random.seed(seed)
         dfs = []
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import logging
+import random
 import warnings
 from pathlib import Path
 from typing import Any, Callable, Iterable, Literal, Type, cast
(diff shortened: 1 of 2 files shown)


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -240,6 +240,82 @@ def head(self, n: int = 5) -> npd.NestedFrame:
...
+        Args:
+            partition_id (int): the partition to sample.
+            n (int): the number of desired rows.
+
+        Returns:
+            A pandas DataFrame with up to `n` rows of data.
+        """
+        dfs = []
+        remaining_rows = n
+        partition = self._ddf.partitions[partition_id]
```
camposandro, line 277: Can we check that `partition_id` is not out of bounds?

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -240,6 +240,82 @@ def head(self, n: int = 5) -> npd.NestedFrame:
...
+        # Get the count of rows in the partition.
+        pixel_rows = len(partition)
+        fraction = n / pixel_rows
+        print(f"Getting {n} / {pixel_rows} = {fraction}")
+        dfs.append(partition.sample(frac=fraction).compute())
+        if len(dfs) > 0:
+            return npd.NestedFrame(pd.concat(dfs))
```
camposandro, line 284: If we're loading samples from a single partition, we would only have a single data frame. Can we just return the result of the previous compute call?

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -240,6 +240,82 @@ def head(self, n: int = 5) -> npd.NestedFrame:
...
+        return self._ddf._meta
+
+    def random_sample(self, n: int = 5) -> npd.NestedFrame:
```
camposandro, line 287: Should we allow setting a seed on the methods that call [Dask DDF sample](https://docs.dask.org/en/latest/generated/dask.dataframe.DataFrame.sample.html), for reproducibility purposes?

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -240,6 +242,105 @@ def head(self, n: int = 5) -> npd.NestedFrame:
...
+            partition_id (int): the partition to sample.
+            n (int): the number of desired rows.
+            seed (int): random seed
+
+        The `seed` argument is passed directly to `random.seed` in order
+        to assist with creating predictable outputs when wanted, such
+        as in unit tests.
+
+        Returns:
+            A pandas DataFrame with up to `n` rows of data.
```
camposandro, line 280:
```suggestion
            A NestedFrame with up to `n` rows of data.
```

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -240,6 +242,105 @@ def head(self, n: int = 5) -> npd.NestedFrame:
...
+        Args:
+            n (int): the number of desired rows.
+            seed (int): random seed
+
+        The `seed` argument is passed directly to `random.seed` in order
+        to assist with creating predictable outputs when wanted, such
+        as in unit tests.
+
+        Returns:
+            A pandas DataFrame with up to `n` rows of data.
```
camposandro, line 312:
```suggestion
            A NestedFrame with up to `n` rows of data.
```

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -240,6 +242,105 @@ def head(self, n: int = 5) -> npd.NestedFrame:
...
         return self._ddf._meta
 
+    def tail(self, n: int = 5) -> npd.NestedFrame:
+        """Returns a few rows of data from the end of the catalog for previewing purposes.
+
+        Args:
+            n (int): The number of desired rows.
+
+        Returns:
+            A pandas DataFrame with up to `n` rows of data.
```
camposandro, line 252:
```suggestion
            A NestedFrame with up to `n` rows of data.
```

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -219,13 +221,13 @@ def partitions(self):
...
     def head(self, n: int = 5) -> npd.NestedFrame:
-        """Returns a few rows of data for previewing purposes.
+        """Returns a few rows of initial data for previewing purposes.
 
         Args:
             n (int): The number of desired rows.
 
         Returns:
-            A pandas DataFrame with up to `n` of data.
+            A pandas DataFrame with up to `n` rows of data.
```
camposandro, line 230:
```suggestion
            A NestedFrame with up to `n` rows of data.
```

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -240,6 +242,105 @@ def head(self, n: int = 5) -> npd.NestedFrame:
...
+            zip(rows_per_partition, self._ddf.partitions, row_counts)
+        ):
+            if not rows:
+                continue
+            fraction = rows / part_rows
+            logging.debug("Sampling %d / %d rows from partition %d", rows, part_rows, i)
+            selection = part.sample(frac=fraction)
+            dfs += selection.to_delayed()
+        if len(dfs) > 0:
+            return npd.NestedFrame(pd.concat(dask.compute(*dfs)))
```
camposandro, line 341: Here it is likely more efficient to do `dd.concat(dfs).compute()` because Dask might optimize the construction of the task graph for the concatenation. With the current line I think we might need to resolve all the partitions and have them in memory at once.

gitosaurus: Good point about the parallel resolution of the partitions. In terms of memory usage, that should be fine, as the user will often be asking for a relatively small (100 rows at most) amount of data. I definitely do want the `.compute` calls to happen in parallel, and was trusting that the outer `Client` would see to that.

I like the idea of letting Dask optimize the graph. When I tried that line, however, I got this error from `dd.concat`:

```
Input must be a Pandas DataFrame or Series.`
```

which did surprise me. Maybe I need to `dd.concat` instead of calling `.to_delayed`, above? Will try.

gitosaurus: I did try this, calling `dask.compute(dd.concat(dfs))` where `dfs` was the list of sampled partitions. To my surprise, even when the length of `dfs` was only 2, I could see (with debug logging enabled) that Dask was downloading *every* partition to the client, taking 20m before I finally canceled it. In contrast, the existing implementation takes 15s/partition, or ~30s for two partitions. I'll leave the existing implementation for now.

Looking at the [documentation](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.compute.html), it seemed possible that setting `concatenate=False` might help, but it didn't.

For reference, this is what I tried:
```python
        for i, (rows, part, part_rows) in enumerate(
            zip(rows_per_partition, self._ddf.partitions, row_counts)
        ):
            if not rows:
                continue
            fraction = rows / part_rows
            logging.debug("Sampling %d / %d rows from partition %d", rows, part_rows, i)
            selection = part.sample(frac=fraction)
            dfs.append(selection)
        if len(dfs) > 0:
            # Permit Dask to optimize the task graph of the computation.
            return npd.NestedFrame(dask.compute(dd.concat(dfs)), concatenate=False)
        return self._ddf._meta
```

camposandro: Thanks Derek for testing it out - this was a valuable finding!

camposandro (approved):

PR #699, opened 2025-04-10
Expand possible pixelsearch inputs

### LINKED ISSUE #658: Need a compact, easy method for getting pixel from (ra, dec) coordinates
**Feature request**

When a user wants to to select a position in the sky and then select all the objects that are in the same HATS pixel, this can be done easily via this method:

```python
mycatalog.get_partition(order=my_order, pixel=my_pixel)
```

but this assumes that you already know which (order, pixel) pair contains your (ra, dec) coordinate.

One method of accomplishing this today is to do a cone- or box-search on the area of the sky you want, and be sure to include the columns Norder and Npix.  Assuming that the HATS pixel you're trying to find is the one that contains (ra=280, dec=-60), this will do it: 

```python
lsdb.read_hats(
    'https://data.lsdb.io/hats/gaia_dr3/gaia',
    columns=['ra', 'dec', 'Norder', 'Npix'],
    search_filter=lsdb.ConeSearch(ra=280, dec=-60, radius_arcsec=30)).compute()
```

However, `Norder` and `Npix` columns will not be in catalogs going forward, so this isn't a trustworthy recipe.

Another approach is to use the MOC search:

```python
import astropy.units as u
from mocpy import MOC

moc = MOC.from_lonlat(280*u.deg, -60*u.deg, max_norder=29) # 29 is the order for _healpix_29

gaia.moc_search(moc) # returns a catalog with the desired partition (order 4, Npix 2947)
```

While the above works, it's not as intuitive.  What's desired is some top-level mapping function that would look something like:

```python
hpixel = gaia3.healpix_pixel_for_point(ra=280, dec=-60)
gaia3.get_partition(order=hpixel.order, pixel=hpixel.pixel)
```

that would return a `HealpixPixel` with the (order, pixel) pair that can be passed directly to `.get_partition`, or perhaps even something more direct:

```python
gaia3.get_partition_for_point(ra=380, dec=-60)
```

**Before submitting**
Please check the following:

delucchi-cmu:
Closes #658


src/lsdb/core/search/pixel_search.py
```diff
@@ -18,9 +19,30 @@ class PixelSearch(AbstractSearch):
     Does not filter points inside those partitions.
     """
 
-    def __init__(self, pixels: list[tuple[int, int]]):
+    def __init__(self, pixels: tuple[int, int] | HealpixPixel | list[tuple[int, int] | HealpixPixel]):
         super().__init__(fine=False)
-        self.pixels = [HealpixPixel(o, p) for o, p in set(pixels)]
+        if isinstance(pixels, tuple):
```
gitosaurus, line 24:
```suggestion
        if isinstance(pixels, tuple) and len(pixels) == 2:
```

src/lsdb/core/search/pixel_search.py
```diff
@@ -18,9 +19,30 @@ class PixelSearch(AbstractSearch):
...
+            self.pixels = [HealpixPixel(pixels[0], pixels[1])]
+        elif isinstance(pixels, HealpixPixel):
+            self.pixels = [pixels]
+        elif pd.api.types.is_list_like(pixels):
+            if len(pixels) == 0:
+                raise ValueError("Some pixels required for PixelSearch")
+            self.pixels = list({get_healpix_pixel(pix) for pix in pixels})
```
gitosaurus, line 31: Why cast through `set` on the way to `list`? Duplicates are expected?

src/lsdb/core/search/pixel_search.py
```diff
@@ -18,9 +19,30 @@ class PixelSearch(AbstractSearch):
     Does not filter points inside those partitions.
     """
 
-    def __init__(self, pixels: list[tuple[int, int]]):
+    def __init__(self, pixels: tuple[int, int] | HealpixPixel | list[tuple[int, int] | HealpixPixel]):
         super().__init__(fine=False)
-        self.pixels = [HealpixPixel(o, p) for o, p in set(pixels)]
+        if isinstance(pixels, tuple):
+            self.pixels = [HealpixPixel(pixels[0], pixels[1])]
```
gitosaurus, line 25:
```suggestion
            self.pixels = [get_healpix_pixel(pixels)]
```
Just for consistency with what happens with lists of pixels.

gitosaurus (approved):
Looks good! My only comments are optional and mostly stylistic.

src/lsdb/core/search/pixel_search.py
delucchi-cmu: `get_healpix_pixel` performs this check, so changing the below line takes care of it.

src/lsdb/core/search/pixel_search.py
delucchi-cmu: It was restricting to set previously, so I maintained the behavior, but it doesn't really matter for the `filter_from_pixel_list` call. Removed.

PR #700, opened 2025-04-10
Address mypy issues with latest numpy

delucchi-cmu:
Numpy 2.2.0 release cites "Many improved annotations." that are breaking our type checking!@!

Additionally makes `lsdb.from_dataframe` consistent with hats-import pipeline, with changes from https://github.com/astronomy-commons/hats-import/pull/512


camposandro (approved):
Thanks, looks good!

Issue #689, opened 2025-04-05
Single row retrieval method using index catalog


nevencaplar:
Revisit index search method. Assume that index catalog in catalog collections is already avaliable. The function should be `id_search(id_value, id_col=”name_of_id_collumn”)`. If the `id_col` is not supplied query the default id column. 

For instance 
`id_search(id_value)` give a row with object_id = id_value
`id_search(id_value, id_col = "source_id")` gives a row with source_id = id_value

If no index catalog is avaliable is avaliable in the catalog collection fail gracefully.

hombit:
After talking with the ZTF ML (Scope project) team yesterday, and discussing with @delucchi-cmu, @smcguire-cmu, and @camposandro, I propose to slightly change the focus of this feature request to support value matching on any column, rather than limiting it to ID-like searches only.

The proposed interface is `Catalog.match_values(col1=val1, col2=val2)`.

1. I propose not to use the notion of a default ID column, as it is too implicit and we do not provide users with a way to determine which column is the default one. Good examples are Rubin DR source tables (objectId or sourceId), [SDSS photometric catalogs](https://vizier.cds.unistra.fr/viz-bin/VizieR?-source=V/154&-to=3) (objID, SDSS16, or SDSS-ID), PS1 DRs (objName or objID, objName is what you normally see in papers), and VSX (OID or Name, Name is commonly used by variable star people).

2. Index catalog search may also cover other use cases, e.g. retrieving multiple rows based on the value of a diverse flag or a text column. For example, the VSX "Type" column has 1,670 different values for 11 million objects. Another example is querying based on observational field or sky patch.

3. https://github.com/astronomy-commons/lsdb/issues/646 proposes an interface for retrieving objects by patch and tract. The same was asked by the Scope people for ZTF's `field` and `ccdid`. I believe we can absolutely support this via index catalogs, if we allow a composite index. This would make it possible to do `ztf.match_values(fieldid=217, ccdid=13)`.

4. If an index catalog is not available for a given combination of fields, I propose returning an error message that:
   a) lists all available index catalogs (or provides a method suggestion to list them), and  
   b) suggests using `Catalog.query("col={value}")` or `Catalog[Catalog[col] == value]` instead.

delucchi-cmu:
Suggestion from live discussion:

1) keep name `id_search`, but use the `match_values` behavior where keys on kwargs are the column names.
2) remove notion of a default index from method.
[...]

PR #710, opened 2025-04-21
Add catalog `id_search` 

camposandro:
Remove `Catalog.index_search` in favor of `Catalog.id_search`. This new method leverages the collections' index information, but it also supports specifying index catalogs per field directly as an argument to the function. This is useful for fields that lack an index in the collection, or when working with standalone catalogs.

The interface follows the suggestions proposed in [astronomy-commons/lsdb#689 (comment)](https://github.com/astronomy-commons/lsdb/issues/689#issuecomment-2813860706), with one small deviation: field-value pairs are passed as a dictionary rather than as keyword arguments, allowing the use of field names that are not valid Python identifiers (e.g. with spaces, dashes, etc.).


src/lsdb/core/search/index_search.py
```diff
@@ -1,34 +1,49 @@
...
+    index_catalogs: dict[str, HCIndexCatalog]
+    """Mapping of field name to respective index catalog"""
 
-    def __init__(self, ids, catalog_index: IndexCatalog, fine: bool = True):
-        super().__init__(fine)
-        self.ids = ids
-        self.catalog_index = catalog_index
+    def __init__(self, values: dict[str, Any], index_catalogs: dict[str, HCIndexCatalog]):
+        super().__init__(fine=True)
+        if values.keys() != index_catalogs.keys():
```
delucchi-cmu, line 24: This feels a bit strong. I can imagine someone having some canned dictionary of index catalogs, but not always using all of them. Could just check that `np.all(values.keys in index_catalog.keys())` or however that's written.

src/lsdb/core/search/index_search.py
```diff
@@ -1,34 +1,49 @@
...
-    """
+    index_catalogs: dict[str, HCIndexCatalog]
+    """Mapping of field name to respective index catalog"""
 
-    def __init__(self, ids, catalog_index: IndexCatalog, fine: bool = True):
-        super().__init__(fine)
-        self.ids = ids
-        self.catalog_index = catalog_index
+    def __init__(self, values: dict[str, Any], index_catalogs: dict[str, HCIndexCatalog]):
+        super().__init__(fine=True)
```
delucchi-cmu, line 23: I think we should still allow `fine=False`, for a case like "I want the partition with my host galaxy so I can look at other nearby objects".

src/lsdb/core/search/index_search.py
```diff
@@ -1,34 +1,49 @@
...
+            )
+        self.values = values
+        self.index_catalogs = index_catalogs
 
     def filter_hc_catalog(self, hc_structure: HCCatalogTypeVar) -> HCCatalogTypeVar:
-        healpix_pixels = self.catalog_index.loc_partitions(self.ids)
-        return hc_structure.filter_from_pixel_list(healpix_pixels)
+        """Determine the pixels for which there is a result in each field"""
+        all_pixels = set(hc_structure.get_healpix_pixels())
+        for field_name, field_value in self.values.items():
```
delucchi-cmu, line 36: We're going to have to think more about what it would look like for the case of a complex index, like tract+patch. But I think this is ok for a first pass.

camposandro: Yes, this currently only works for simple indexes. I think we do not yet support the generation of hats `IndexCatalog` with composite indexes, but tract+patch should still work with this implementation as long as we have a separate index catalog for each field.

delucchi-cmu (approved):

PR #713, opened 2025-04-22
Migrate Nested-Dask into LSDB.nested

dougbrn:
Removes dependency on nested-dask, preserves all of nested-dask in the lsdb.nested module. Swapping out `import nested_dask as nd` with `import lsdb.nested as nd` is the only access change to this codebase.

Added in-line annotations to functions that are either not wrapped, fully unused, or just a bit different between LSDB and nested-dask. Will create a tracker issue for these as well.


src/lsdb/catalog/association_catalog.py
```diff
@@ -1,5 +1,5 @@
 import hats as hc
-import nested_dask as nd
```
delucchi-cmu, line 2: LSDB uses pylint/isort instead of ruff, so just about all of the imports will need to be re-ordered. Should be a pretty painless `isort .`

src/lsdb/nested/backends.py
```diff
@@ -0,0 +1,46 @@
+# Python 3.9 doesn't support "|" for types
```
delucchi-cmu, line 1: LSDB doesn't support python 3.9, though! No need to address in this PR - can be an easy follow-up.

src/lsdb/nested/datasets/generation.py
```diff
@@ -0,0 +1,74 @@
+from nested_pandas import datasets
+
+import lsdb.nested as nd
+
+
+def generate_data(n_base, n_layer, npartitions=1, seed=None) -> nd.NestedFrame:
```
delucchi-cmu, line 6: I'd really like this to generate data with radec values, so it can be made into a catalog.

But again - can be in a follow-up.

delucchi-cmu (approved):
Please address pre-commits - otherwise LGTM!

src/lsdb/nested/datasets/generation.py
dougbrn: Will make an issue!

src/lsdb/nested/backends.py
dougbrn: Will make an issue!

src/lsdb/catalog/association_catalog.py
dougbrn: Gotcha, made pre-commit happy

PR #719, opened 2025-04-24
Don't flatten numpy array with unique call.

### LINKED ISSUE #718: numpy failure with all-sky from_dataframe
**Bug report**

Based on issue from Sam Charney, we have a problem creating a catalog using `from_dataframe` using `drop_empty_siblings=True` when the input catalog covers the whole sky.

I'm working on a repro, but it looks like it's a subtle object vs number issue in our numpy unique logic. Who knew!??!

**Before submitting**
Please check the following:

delucchi-cmu:
This was kind of a pain to track down!

The call to `np.unique` will flatten the `alignment` array when there are NO `None` values (but only then!). I've changed to use a more explicit construction of the desired pixel list.

The reproducer CSV comes from taking just the first 11 rows from each of the partitions from Sam Charney's provided catalog's coordinates. If I only took 10 rows per partition, it didn't trigger the error. I'm at a loss on that one, but I need to move on with my day.

Closes #718


src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -197,9 +197,7 @@ def _compute_pixel_list(self) -> list[HealpixPixel]:
             threshold=self.threshold,
             drop_empty_siblings=self.drop_empty_siblings,
         )
-        non_none_elements = alignment[alignment != np.array(None)]
-        pixel_list = np.unique(non_none_elements)
-        pixel_list = [HealpixPixel(order, pix) for (order, pix, count) in pixel_list if int(count) > 0]
+        pixel_list = list({HealpixPixel(tup[0], tup[1]) for tup in alignment if not tup is None})
```
camposandro, line 200: Here should we still make sure to filter by count, in case `drop_empty_siblings` is False?
```suggestion
        pixel_list = list({HealpixPixel(tup[0], tup[1]) for tup in alignment if not tup is None and int(tup[2]) > 0})
```

delucchi-cmu: No - those will be filtered out by the alignment logic and they will just be `None` instead of a tuple with a 0-value for the third element.

camposandro (approved):
What a nasty bug indeed!

PR #729, opened 2025-04-30
LSDB.nested catalog generation

### LINKED ISSUE #722: Make LSDB.nested dataset generation functions have radec columns
**Feature request**
The generator functions in LSDB.nested.datasets.generation allow quick construction of dummy datasets to play with, these are currently not valid HATS as they lack any radec information. These should be updated to have that radec information so that they are HATS-able.

**Before submitting**
Please check the following:

### LINKED ISSUE #730: LSDB.from_dataframe loses Nested-Pandas dtypes on load
**Bug report**
Uncertain if this is introduced by Nested-Pandas v0.4.0 or not (encountered in working with v0.4.0+), but schema generation  in LSDB.from_dataframe uses pyarrow dtypes, which sheds the NestedDtype since it's pyarrow backed. When going from a nested-pandas NestedFrame to an LSDB catalog we should preserve the NestedDtypes for nested columns.

There's some machinery in nested-pandas to scan pyarrow schema for struct-lists and attempt the cast, so this may be just an exercise in using that. Alternatively, we know what has a nesteddtype from the input nestedframe, so it may just be easiest to scan that and update the dtypes on the resulting catalog accordingly.

**Before submitting**
Please check the following:

dougbrn:
Closes #722 
Closes #730 

Modifies generate_data with ra/dec & and integer id column
Adds generate_catalog one-liner
Allows use of search_filters like in to_hats to control spatial region of generated data. Currently supports BoxSearch (equivalent to just using ra_range and dec_range) and ConeSearch
Removes generate_parquet_file because it's not really relevant to LSDB

These functions are currently found in the import structure as:
```
from lsdb.nested import generate_data, generate_catalog
# or
from lsdb.nested.datasets import generate_data, generate_catalog
```
Is there a better place for these?


dougbrn:
One failed test on the windows side that seems unrelated...

src/lsdb/nested/datasets/generation.py
```diff
@@ -1,9 +1,16 @@
...
 from nested_pandas import datasets
 
+import lsdb
+from lsdb.core.search import BoxSearch, ConeSearch, PixelSearch
 from lsdb.nested.core import NestedFrame
 
 
-def generate_data(n_base, n_layer, npartitions=1, seed=None) -> NestedFrame:
+def generate_data(
+    n_base, n_layer, npartitions=1, seed=None, ra_range=(0.0, 360.0), dec_range=(-90, 90), search_filter=None
```
delucchi-cmu, line 12: Based on #736, can you go ahead and name this `region`?

src/lsdb/nested/datasets/generation.py
```diff
@@ -53,22 +182,39 @@ def generate_parquet_file(n_base, n_layer, path, file_per_layer=True, npartition
...
 
     Returns
     -------
-    None
+    Catalog
+        The constructed LSDB Catalog.
+
+    Examples
+    --------
+    >>> from lsdb.nested.datasets import generate_catalog # doctest: +SKIP
```
delucchi-cmu, line 205: But I would love to use `generate_catalog` in other doc-string-test-things!! Is this slow, or just annoying? or non-deterministic with other machines?

src/lsdb/nested/datasets/generation.py
```diff
@@ -53,22 +182,39 @@ def generate_parquet_file(n_base, n_layer, path, file_per_layer=True, npartition
...
+    >>> from lsdb.core.search import ConeSearch # doctest: +SKIP
+    >>> generate_data(10, 100, search_filter=ConeSearch(5, 5, 1)) # doctest: +SKIP
     """
-    nf = generate_data(n_base, n_layer, npartitions, seed)
-    nf.to_parquet(path, by_layer=file_per_layer, write_index=False)
+
+    base_nf = generate_data(
+        n_base, n_layer, seed=seed, ra_range=ra_range, dec_range=dec_range, search_filter=search_filter
+    )
+    return lsdb.from_dataframe(base_nf.compute(), catalog_name="generated_catalog", **kwargs)
```
delucchi-cmu, line 220: If I pass `catalog_name="fun_times"` as a kwarg, which one wins?

dougbrn: Good catch, updated this to just name it "generated_catalog" if a name isn't provided, but no longer blocks the use of a custom name

src/lsdb/nested/datasets/generation.py
dougbrn: The doctest skipping is a hold-over from the lsdb.nested port,  but these should be fully usable in doctests. Just be mindful of setting a seed when expecting certain output and being aware of how large of a dataset you're trying to load, I've enabled doctests here, but just assign the results to variables to not check against a certain output.

(I did notice doctest and pylint were butting heads a bit on the string dataframe output, so that may come up as well in future doctests)

src/lsdb/nested/datasets/generation.py
dougbrn: Done!

delucchi-cmu (approved):

PR #742, opened 2025-05-05
Support Nested Parquet Serialization

smcguire-cmu:


pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
     "dask[complete]",
     "deprecated",
     "hats>=0.5.1",
-    "nested-pandas>=0.3.8,<0.4.0",
+    "nested-pandas>=0.3.8,<0.5.0",
```
dougbrn, line 26: I think we should probably pin the minimum nested-pandas to 0.4.0 here as well

dougbrn (approved):
Nice, looks good! Just one point to address.

PR #745, opened 2025-05-06
remove 3.9 handling in lsdb.nested backend

### LINKED ISSUE #723: Remove 3.9 handling in lsdb.nested backends
LSDB no longer supports 3.9, so this should be removed: https://github.com/astronomy-commons/lsdb/blob/main/src/lsdb/nested/backends.py#L1C1-L2C35

dougbrn:
Closes #723

### DIFF
--- a/src/lsdb/nested/backends.py
+++ b/src/lsdb/nested/backends.py
@@ -1,6 +1,3 @@
-# Python 3.9 doesn't support "|" for types
-from __future__ import annotations
-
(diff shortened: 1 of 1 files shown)


camposandro (approved):
Looks good to me!

Should we also remove this futures annotation from other modules?

dougbrn:
@camposandro, probably! Not sure where else it's used in the package but it would make sense to move on elsewhere as well.

PR #758, opened 2025-05-08
Add data thumbnail creation to `to_hats`

camposandro:
Create the data thumbnail parquet for `lsdb.Catalogs` when saving them to disk with `to_hats`.

### DIFF
--- a/src/lsdb/io/to_hats.py
+++ b/src/lsdb/io/to_hats.py
@@ -94,7 +94,8 @@ def to_hats(
         overwrite (bool): If True existing catalog is overwritten
         create_thumbnail (bool): If True, create a data thumbnail of the catalog for previewing purposes.
             Defaults to False.
-        thumbnail_threshold (int): The number of points per pixel as specified at import time.
+        thumbnail_threshold (int): The maximum number of rows in the data thumbnail,
+            which is otherwise one row/partition.
             Defaults to 1_000_000.
         **kwargs: Arguments to pass to the parquet write operations
     """
--- a/src/lsdb/io/to_hats.py
+++ b/src/lsdb/io/to_hats.py
@@ -75,6 +75,8 @@ def to_hats(
     default_columns: list[str] | None = None,
     histogram_order: int = 8,
     overwrite: bool = False,
+    create_thumbnail: bool = False,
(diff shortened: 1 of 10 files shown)


src/lsdb/io/to_hats.py
```diff
@@ -90,6 +92,10 @@ def to_hats(
             exist.
         histogram_order (int): The default order for the count histogram. Defaults to 8.
         overwrite (bool): If True existing catalog is overwritten
+        create_thumbnail (bool): If True, create a data thumbnail of the catalog for previewing purposes.
+            Defaults to False.
+        thumbnail_threshold (int): The number of points per pixel as specified at import time.
```
gitosaurus, line 97:
```suggestion
        thumbnail_threshold (int): The maximum number of rows in the data thumbnail,
            which is otherwise one row/partition.
```

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -81,7 +81,11 @@ def __init__(
         self.drop_empty_siblings = drop_empty_siblings
         self.threshold = self._calculate_threshold(partition_size, threshold)
         self.catalog_info = self._create_catalog_info(
-            ra_column=ra_column, dec_column=dec_column, total_rows=int(len(self.dataframe)), **kwargs
+            ra_column=ra_column,
+            dec_column=dec_column,
+            total_rows=int(len(self.dataframe)),
```
smcguire-cmu, line 86: Why the int cast here? Numpy?

camposandro: Thanks, it shouldn't be needed. I'm not sure why it was there before!

src/lsdb/io/to_hats.py
```diff
@@ -90,6 +92,10 @@ def to_hats(
...
+            Defaults to 1_000_000.
```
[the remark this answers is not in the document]
camposandro: Thanks @gitosaurus! When we write the partitions out to disk we indeed get the counts for each of them. We can use the max value for those pixel counts as our `hats_max_rows`.

Issue #765, opened 2025-05-14
Catalog `all_columns` property should be same as `columns` for unserialized catalogs


gitosaurus:
**Bug report**

The `all_columns` property is intended to return the complete list of columns available on disk, distinguished from `columns`, which are the names of the columns which have been loaded into memory.

But when a catalog has never been serialized to disk, this property should be identical to `columns`, as it has the same meaning.  Today, however, an attempt to use the property results in this case results in a `ValueError`.

One such case is cross-matching.  If you do `comcam_x_milliquas = comcam_cat.crossmatch(miliquas_cat)`, it runs fine, but trying to look at comcam_x_milliquas creates a huge error message which includes this traceback:

```
---> 35 available_cols = len(self.all_columns)
     36 return (
     37     f"<div><strong>lsdb Catalog {self.name}:</strong></div>"
     38     f"{data}"
     39     f"<div>{loaded_cols} out of {available_cols} columns in the catalog have been loaded "
     40     f"<strong>lazily</strong>, meaning no data has been read, only the catalog schema</div>"
     41 )

File ~/.local/lib/python3.12/site-packages/lsdb/catalog/dataset/dataset.py:83, in Dataset.all_columns(self)
     81 """Returns all columns in the original Dataset"""
     82 if self.hc_structure.original_schema is None:
---> 83     raise ValueError("Original Catalog Columns are not available")
     84 col_names = self.hc_structure.original_schema.names
     85 if SPATIAL_INDEX_COLUMN in col_names:

ValueError: Original Catalog Columns are not available
```

Instead, the lack of `self.hc_structure.original_schema` should indicate that `all_columns` is identical to `columns`, and should return that data.

**Before submitting**
Please check the following:

- [X] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [X] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.
- [X] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

gitosaurus:
Closed by https://github.com/astronomy-commons/lsdb/pull/766

PR #766, opened 2025-05-14
Raise a helpful error message when the user is accessing a catalog's column that hasn't been loaded.

### LINKED ISSUE #685: Raise error when accessing column that hasn't been loaded
Raise a helpful error message when the user is accessing a catalog's column that hasn't been loaded. We can make use of the `all_columns` property introduced in #682 to verify it.

gitosaurus:
Now that we have default columns for catalogs, there are usually columns available on disk that aren't pulled into memory, and this is even more true when the `columns=` argument is used.  When the user tries to reference columns that are part of the catalog but which weren't loaded, the error messages imply that the columns were never available.  Instead, building on the `all_columns` property, give a more specific error when the user tries to access an unloaded column.

Closes #685 and #765 .

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -150,6 +150,7 @@ class Catalog(HealpixDataset):
                 new_data = da.arange(...)
                 catalog = catalog.assign(new_col=new_data)
         """
+        self.check_unloaded_columns(list(kwargs.keys()))
         ddf = self._ddf.assign(**kwargs)
         return self._create_updated_dataset(ddf=ddf)
 
@@ -503,6 +504,7 @@ class Catalog(HealpixDataset):
         Returns:
             A new Catalog containing the results of the column match.
         """
+        self.check_unloaded_columns(list(values.keys()))
 
         def _get_index_catalog_for_field(field: str):
             """Find the index catalog for `field`. Index catalogs declared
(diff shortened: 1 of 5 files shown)


src/lsdb/catalog/catalog.py
```diff
@@ -677,6 +679,21 @@ def merge(
             suffixes = (f"_{self.name}", f"_{other.name}")
         if len(suffixes) != 2:
             raise ValueError("`suffixes` must be a tuple with two strings")
+
+        # This is kind of a mess
+        def harmonize(col: str | list[str] | None) -> list[str]:
+            if not col:
```
smcguire-cmu, line 685: I think it's better to explicitly check `is not None` here. I think an empty string would trigger this clause which I don't think is intended behavior.

src/lsdb/catalog/dataset/dataset.py
```diff
@@ -92,3 +97,22 @@ def original_schema(self):
         if self.hc_structure.original_schema is None:
             raise ValueError("Original Catalog Columns are not available")
         return self.hc_structure.original_schema
+
+    def check_unloaded_columns(self, column_names: Sequence[str | None] | None):
```
smcguire-cmu, line 101: I would make this a private method with an underscore, but I could be convinced this is better.

gitosaurus: I agree.  Always easier to go from private to public later if need be.  Will fix.

smcguire-cmu (approved):

PR #768, opened 2025-05-15
Support only providing list_columns in catalog.nest_lists

### LINKED ISSUE #583: Simplify interface of nest_lists
**Feature request**

Current interface asks to set `base_columns` as mandatory argument, which is very annoying for the cases when I have few list columns only, see an example of frustration here, list comprehensions:
https://docs.lsdb.io/en/stable/tutorials/pre_executed/ztf-alerts-sne.html#Load-catalog-structure-and-%22nest%22-columns

It would be nice to make `base_column` optional and allow to specify `list_columns` only

**Before submitting**
Please check the following:

- [ ] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.

wilsonbb:
Fixes #583 

Modifies the lsdb wrappers around `NestedFrame.from_lists` so they match the nested-pandas behavior of allowing the `base_columns` to be optional. When`base_columns` are not provided, all non-list columns are inferred to be based columns. This required no changes from LSDB outside of the function signatures of the wrappers.

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -870,7 +870,8 @@ class Catalog(HealpixDataset):
 
         Args:
             base_columns (list-like or None): Any columns that have non-list values in the input catalog.
-                These will simply be kept as identical columns in the result
+                These will simply be kept as identical columns in the result. If None, is inferred to be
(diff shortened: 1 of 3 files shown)


hombit (approved):
Thank you

Issue #795, opened 2025-05-30
Populated join partition correctly for association


delucchi-cmu:
**Bug report**

Currently, the `join_Norder` and `join_Npix` are just blindly copied from the left side's order/pixel. Use the `_healpix_29` from the right, as well as partitioning, to appropriately populate this pixel value.

**Before submitting**
Please check the following:

- [ ] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [ ] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.
- [ ] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

camposandro:
Closing this issue since we decided the partition join info file is to be removed.

PR #811, opened 2025-06-05
Join through association

camposandro:
The association tables may need to be repartitioned such that its pixels are different from the inner alignment between the primary and the join catalogs. If that's the case we need to recompute the alignment for the join. 

This PR adds some logic to do that: `align_catalog_with_association` first aligns the primary catalog with the association catalog and then it aligns the result to the join catalog to obtain the pixels that need to be joined between the three. 

This is a step toward https://github.com/astronomy-commons/lsdb/issues/795. However, I am looking for feedback since this solution will introduce new `assoc_Norder` and `assoc_Npix` columns in the partition join info (apart from the already existing Norder, Npix and join_Norder, join_Npix).

### DIFF
--- a/src/lsdb/dask/merge_catalog_functions.py
+++ b/src/lsdb/dask/merge_catalog_functions.py
@@ -19,7 +19,6 @@ from hats.pixel_tree.pixel_alignment import align_with_mocs
 from hats.pixel_tree.pixel_tree import PixelTree
 from mocpy import MOC
 
-import lsdb
 import lsdb.nested as nd
 from lsdb.dask.divisions import get_pixels_divisions
 from lsdb.types import DaskDFPixelMap
@@ -28,6 +27,10 @@ if TYPE_CHECKING:
     from lsdb.catalog.association_catalog import AssociationCatalog
     from lsdb.catalog.catalog import Catalog
     from lsdb.catalog.dataset.healpix_dataset import HealpixDataset
+
+
+ASSOC_NORDER = "assoc_Norder"
+ASSOC_NPIX = "assoc_Npix"
 
 
 def concat_partition_and_margin(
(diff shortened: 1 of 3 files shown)


src/lsdb/dask/merge_catalog_functions.py
```diff
@@ -71,7 +76,60 @@ def align_catalogs(left: Catalog, right: Catalog, add_right_margin: bool = True)
...
+    return align_with_mocs(
+        left.hc_structure.pixel_tree,
+        right_tree,
+        left.hc_structure.moc,
+        right_moc,
+        alignment_type=PixelAlignmentType.INNER,
+    )
+
 
+def align_catalogs_with_association(
```
delucchi-cmu, line 89: After discussing this morning, this approach makes even more sense, and I like it.

src/lsdb/io/to_association.py
```diff
@@ -22,16 +22,17 @@ def perform_write(
     df: npd.NestedFrame,
     hp_pixel: HealpixPixel,
     base_catalog_dir: str | Path | UPath,
+    separation_column: str,
```
smcguire-cmu, line 25: Not all crossmatch algorithms are required to give a separation column, and if they do it might be in different units. And it's annoying for the user to have to provide this. I think it will be easier to just compute the max separation from the ra dec columns.

camposandro: Thanks, that is a very good point!

I was getting around to making this change, but I realized it's not as intuitive as I expected, since we currently don't keep track of the properties of the catalogs involved in a crossmatch. From an API perspective, I'm also not sure what that should look like either. I opened issue https://github.com/astronomy-commons/lsdb/issues/945 to hopefully tackle this as an improvement.

delucchi-cmu (approved):

PR #821, opened 2025-06-11
Include skymap in to_hats

### LINKED ISSUE #820: Write skymap.fits on to_hats
The point_map.fits file is being replaced by `skymap.fits` (and similar files at other orders). The `to_hats` method should write out this file.

With this change, also confirm that the `hats.properties` file is written alongside the `properties` file.

delucchi-cmu:
Closes #820


camposandro (approved):
Looks good!

PR #828, opened 2025-06-12
include ra and dec columns

### LINKED ISSUE #556: Include ra/dec in column list, even if not in `default_columns`
**Bug report**

If a catalog has metadata for `default_columns`, we will only load those by default. If however, that list of columns does not contain the proper ra/dec columns, those will not be loaded. This creates errors on crossmatch that are unexpected.

We should check the default columns to make sure they have radec columns, and add them if not present.

**Before submitting**
Please check the following:

kesiavino:
include ra and dec columns if they're not in the specified columns or default_columns when loading a catalog

Closes #556


camposandro:
Thank you! Can you please just run the formatter before merging?

Issue #767, opened 2025-05-14
Improve API reference documentation


hombit:
API reference documentation is hard to navigate and use. A few possible improvements:

- [x] Remove prefixes from function/method names, as done in the [`nested-pandas` docs](https://nested-pandas.readthedocs.io/en/latest/reference/nestedframe.html).
- [ ] Add hyperlinks for `hats`/`LSDB` functions, methods, and classes mentioned and used in the docs, including "Getting Started" and "Tutorials". See the [`astropy`](https://docs.astropy.org/en/stable/units/index.html) and [`tinygp`](https://tinygp.readthedocs.io/en/stable/tutorials/intro.html#sampling) docs.
- [x] Improve and clean up the structure: there is no `Catalog` on the autoapi index page, the `Catalog` page has the `hc_structure` attribute mentioned twice, and a search for "crossmatch" lists `Catalog.crossmatch` in the 6th position.

PR #853, opened 2025-06-26
Mimic reference API docs of nested-pandas

delucchi-cmu:
See #767 and https://github.com/lincc-frameworks/nested-pandas/pull/229


docs/reference/io.rst
```diff
@@ -0,0 +1,21 @@
+==================
+Input/Ouput
```
gitosaurus, line 2:
```suggestion
Input/Output
```

docs/reference/catalog.rst
```diff
@@ -0,0 +1,96 @@
...
+Properties
+~~~~~~~~~~~~~~~~~~
+.. autosummary::
+    :toctree: api/
+
+    Catalog.hc_structure
+    Catalog.dtypes
+    Catalog.columns
+    Catalog.nested_columns
+    Catalog.all_columns
```
gitosaurus, line 24: Welcome, much-used property!  Will this mean that I can refer to it as
```reStructuredText
:attr:`Catalog.all_columns`
```
? That would be lovely!

gitosaurus (approved):
LGTM but check the one suggestion!

docs/reference/catalog.rst
delucchi-cmu: I think you might need to use the full(-er) path of `lsdb.catalog.Catalog.all_columns`, since the `lsdb.catalog` bit is in the `currentmodule` attribute above for convenience.

PR #863, opened 2025-06-27
Allow any column in include_columns

delucchi-cmu:
I would like to find the statistics of a single column, regardless of the loaded default columns. Since this method will only look at the `_metadata` file and no other data partitions, the set of initially-requested columns is irrelevant.


gitosaurus (approved):
Makes sense!

PR #864, opened 2025-06-27
fix bad toc conflict resolution

dougbrn:
Accidentally removed the link to the photo-z nb

### DIFF
--- a/docs/tutorial_toc/toc_rubin.rst
+++ b/docs/tutorial_toc/toc_rubin.rst
@@ -6,5 +6,6 @@ Working with Rubin Data
 
     Accessing Rubin Data Preview 1 </tutorials/pre_executed/rubin_dp1>
     Intro to Rubin Catalog Operations </tutorials/pre_executed/using_rubin_data>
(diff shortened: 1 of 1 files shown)


camposandro (approved):
Thanks :)

docs/tutorial_toc/toc_rubin.rst
```diff
@@ -6,5 +6,6 @@ Working with Rubin Data
 
     Accessing Rubin Data Preview 1 </tutorials/pre_executed/rubin_dp1>
     Intro to Rubin Catalog Operations </tutorials/pre_executed/using_rubin_data>
+    Accessing Rubin Data Preview 1 photo-z </tutorials/pre_executed/rubin_dp1_photoz>
```
camposandro, line 9: @dougbrn Maybe add it after your "visualizing periodic lightcurves" instead?

camposandro (approved):

docs/tutorial_toc/toc_rubin.rst
dougbrn: No strong opinions, so inclined to keep as is unless you feel it would be better :)

PR #884, opened 2025-07-03
partition_size is in rows

gitosaurus:


nevencaplar (approved):
Still worried how this mistake happen, but good fix.

gitosaurus:
It looks to be that it happened because partition size *is* in bytes for
another class related to DataFrame. But it's deep in kwargs and may not
mean the same as an LSDB partition.

PR #907, opened 2025-07-14
Change is_builtin_algorithm to accept enum variants

hombit:
It should fix problems with `crossmatch(radius_arcsec=larger_than_five)`


camposandro (approved):
Thank you, Kostya!

PR #919, opened 2025-07-16
Add option to input a list of values to id_search

### LINKED ISSUE #764: Vectorize index column lookups
The current implementation of `Catalog.id_search` allows for fast multi-index column lookups:

```
# assuming there are index catalogs for both fields
# values for field_id and cc_id cannot be lists
catalog.id_search(values={"field_id": 810, "cc_id": 10000})
```

One of the most popular use cases is fast-object lookup by object ID. It would then be convenient to be able to pass lists of values for the object ID field. Something similar to:

```
catalog.id_search(values={"objid": [1,2,3...])
```

This multi-value column lookups on an index column is not currently supported but it should be allowed if (and only if) we're querying for a single column.

I think we can either:

1) Have the `id_search` method handle both cases (it allowing for only one iterable index column).

2) Rename the current `id_search` to `column_search`, keeping its current implementation to support multi-single value column lookups; and then have another named `id_search` which would provide an easy and intuitive way of querying for ID columns with single values or lists as described.

olivialynn:
Closes #764

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -499,7 +499,10 @@ class Catalog(HealpixDataset):
         precedence over the catalogs specified in the collection.
 
         Args:
-            values (dict[str, Any]): The mapping of field names (as string) to their values.
+            values (dict[str, Any]): The mapping of field names (as string) to their search values.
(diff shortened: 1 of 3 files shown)


delucchi-cmu (approved):
Looks great - thanks!

PR #920, opened 2025-07-16
Add error message when non-catalog used as xmatch target in Catalog.crossmatch

### LINKED ISSUE #816: Improve error message for use of Catalog.crossmatch with non-Catalogs
Right now, calling `hats_catalog.crossmatch(not_hats_catalog)` fails without a clear explanation. This is tricky because we now support crossmatching dataframes and NestedFrames, but we only support that through `lsdb.crossmtach(frame_or_catalog, frame_or_catalog)`.

We should update the method to raise a clearer error when used on a non-Catalog, pointing the user to use lsdb.crossmatch(...) instead.

Would be nice to also comb through documentation and see if relevant notebooks and/or docstrings could be clearer (point to one from the other, or maybe have a little section in a notebook, etc).

olivialynn:
Closes #816


nevencaplar (approved):
LGTM

PR #932, opened 2025-07-21
Source association unique expid

clytieq:
Enforce that each object does not include any sources with the same expid


src/lsdb/core/source_association/baseline_source_associator.py
```diff
@@ -35,36 +34,47 @@ def associate_sources(
...
+                sources_xyz, objects_xyz, n_neighbors=1, max_distance=max_dist_chord
             )
-            if len(object_idx) > 0:
+            # Ensure each object matches to at most one source in this exposure
+            unique_object_idx, object_idx_counts = np.unique(xmatch_objects_idx, return_counts=True)
+            for object_id in unique_object_idx[object_idx_counts > 1]:
+                # Index into crossmatch indices
+                xmatch_idx_idx = np.where(xmatch_objects_idx == object_id)[0]
+                xmatch_idx_idx_dups_idx = np.argsort(xmatch_distances[xmatch_idx_idx])[1:]
+                xmatch_dups_idx_idx = xmatch_idx_idx[xmatch_idx_idx_dups_idx]
```
smcguire-cmu, line 65: Nice job with the numpy logic! It can get tricky. These names are a bit confusing though. I think maybe `xmatch_obj_id_idx`, `xmatch_obj_id_idx_dups_idx`, and `xmatch_idx_dups_idx`, is slightly better, but it's all just numpy stuff that's never going to be too readable. (And you've done a great job improving my variable names in this pr too!).

smcguire-cmu (approved):
Looks good to me! I have one slight suggestion, but great work overall!

PR #933, opened 2025-07-21
scaling_workflows.ipynb: fix a typo

hombit:


dougbrn (approved):
good catch, thank you!

PR #934, opened 2025-07-22
Source association naive aggregator

clytieq:
ensure that basic object aggregator uses ra and dec of source with earliest exposure

changes to the pre-executed juptyer notebook `source_association.ipynb` to test this change will conflict with those in my other pull request for unique expids, but there shouldn't be any changes to any code in the same cells, also they're not super important because I'll eventually want to move those into the tests instead of the tutorial notebook


smcguire-cmu:
Looks like this is trying to merge into main and not the `source-association` branch, please could you change the target branch!

clytieq:
good catch, thanks!

src/lsdb/core/source_association/basic_object_aggregator.py
```diff
@@ -31,9 +37,10 @@ def get_object_meta_df(self) -> npd.NestedFrame:
     def perform_object_aggregation(
         self, df: dict, obj_id: int, pixel: HealpixPixel = None, properties: TableProperties = None
     ) -> dict:
+        idx = np.argmin(df[self.exposure_name])
```
smcguire-cmu, line 40: It's worth doing `.to_numpy` with these pandas series, incase the underlying dtype isn't a numpy array and doesn't play well with numpy.

src/lsdb/core/source_association/basic_object_aggregator.py
```diff
@@ -31,9 +37,10 @@ def get_object_meta_df(self) -> npd.NestedFrame:
...
         return {
             "object_id": obj_id,
-            self.ra_name: df[properties.ra_column][0],
-            self.dec_name: df[properties.dec_column][0],
-            SPATIAL_INDEX_COLUMN: df[SPATIAL_INDEX_COLUMN][0],
+            self.ra_name: df[properties.ra_column][idx],
+            self.dec_name: df[properties.dec_column][idx],
+            SPATIAL_INDEX_COLUMN: df[SPATIAL_INDEX_COLUMN][idx],
```
smcguire-cmu, line 45: I'm just realizing since the idx is an integer from the numpy argmin, it's best to do `.iloc` for accessing these pandas series. This is from my mistake originally, sorry, but could you please fix it here thanks!

PR #980, opened 2025-08-05
Fix catalog html repr when catalog is empty

### LINKED ISSUE #896: Default margin cache created by `lsdb.generate_catalog` has errors when viewed
**Bug report**

```python
from lsdb.nested.datasets import generate_catalog
gen_cat = generate_catalog(1000, 10, ra_range=(0.,10.), dec_range=(-5.,0.))
gen_cat.margin
```
produces:
```
ValueError: Length of values (1) does not match length of index (0)
```

Changing the margin cache parameters to `generate_catalog` (`margin_order=` and `margin_threshold=` doesn't influence the error.

**Before submitting**
Please check the following:

### LINKED ISSUE #927: Error displaying empty catalog
**Bug report**

I'm getting an error when I try to `display()` an empty catalog. Reproducible code to run in Jupyter
```
import lsdb

catalog = lsdb.open_catalog('https://data.lsdb.io/hats/des/des_dr2')
pixel_search = lsdb.PixelSearch.from_radec(80.0, 33.0)
catalog.search(pixel_search)
```

```
ValueError: Length of values (1) does not match length of index (0)
```

**Before submitting**
Please check the following:

smcguire-cmu:
Changes the behavior generating the html representation of an empty catalog by including an index with 1 element how dask expects.

fixes #927 and fixes #896

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -123,6 +123,8 @@ class HealpixDataset(Dataset):
     def _repr_divisions(self):
         pixels = self.get_ordered_healpix_pixels()
         name = f"npartitions={len(pixels)}"
+        # Dask will raise an exception, preventing display,
+        # if the index does not have at least one element.
         if len(pixels) == 0:
             pixels = ["Empty Catalog"]
         divisions = pd.Index(pixels, name=name)


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -121,8 +121,11 @@ def _repr_data(self):
 
     @property
     def _repr_divisions(self):
-        name = f"npartitions={self._ddf.npartitions}"
-        divisions = pd.Index(self.get_ordered_healpix_pixels(), name=name)
+        pixels = self.get_ordered_healpix_pixels()
+        name = f"npartitions={len(pixels)}"
+        if len(pixels) == 0:
+            pixels = ["Empty Catalog"]
```
gitosaurus, line 127:
```suggestion
        # Dask will raise an exception, preventing display,
        # if the index does not have at least one element.
        if len(pixels) == 0:
            pixels = ["Empty Catalog"]
```
I think Dask's expectation is odd enough to warrant a comment in this case.

gitosaurus (approved):
Looks good!

PR #982, opened 2025-08-06
Change `read_pixels` to avoid passing `HATS Catalog` object to dask graph

smcguire-cmu:
In working with the Zubercal dataset, users were running into issues with a memory leak during the `read_pixel` dask task. This led to workflows failing due to running out of memory. This change updates the `read_hats` behavior to avoid passing the `hats` `Catalog` object to the dask task graph, which seems to solve the memory leak problem.

My best guess is that the MOC object that is part of the hats Catalog could be causing this, since we've ran into issues with how they are cloud pickled before and how the Rust interaction works in dask distributed, but further investigation is necessary to confirm this.


delucchi-cmu:
Seconding the concern about scaling - we were originally passing the paths, but moved away from that intentionally, as the string construction was painfully slow for catalogs with 100k+ partitions. The regression in the benchmarks is a real one!

delucchi-cmu (changes requested):
Please update the PR description.

Issue #776, opened 2025-05-19
read_hats `filters` arguments doesn't fully support pyarrow options


smcguire-cmu:
[Parquet filters](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html) support multiple different options for expressing them, including pyarrow expressions, and nested lists of tuples. We currently only support lists of tuples, which means we only support a list of 'or' filters, and you can't specify an 'and' filter, which is done by nesting lists.

Issue #739, opened 2025-05-02
read_hats(filters=None) fails


hombit:
**Bug report**

`filters=None` is the default value of that parameter according to [the pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_parquet.html), but `lsdb.read_hats(filters=None)` fails:

```python
lsdb.read_hats('https://data.lsdb.io/hats/euclid_q1/euclid_q1_merFinalCatalog', margin_cache='https://data.lsdb.io/hats/euclid_q1/euclid_q1_merFinalCatalog_10arcs', filters=None)
```

<details><summary>Traceback</summary>

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 lsdb.read_hats('https://data.lsdb.io/hats/euclid_q1/euclid_q1_merFinalCatalog', margin_cache='https://data.lsdb.io/hats/euclid_q1/euclid_q1_merFinalCatalog_10arcs', filters=None)

File ~/projects/lincc-frameworks/lsdb/src/lsdb/loaders/hats/read_hats.py:114, in read_hats(path, search_filter, columns, margin_cache, dtype_backend, **kwargs)
    112     catalog.hc_collection = hc_catalog  # type: ignore[attr-defined]
    113 else:
--> 114     catalog = _load_catalog(
    115         hc_catalog,
    116         search_filter=search_filter,
    117         columns=columns,
    118         margin_cache=margin_cache,
    119         dtype_backend=dtype_backend,
    120         **kwargs,
    121     )
    122 return catalog

File ~/projects/lincc-frameworks/lsdb/src/lsdb/loaders/hats/read_hats.py:181, in _load_catalog(hc_catalog, search_filter, columns, margin_cache, dtype_backend, **kwargs)
    178 catalog_type = hc_catalog.catalog_info.catalog_type
    180 if catalog_type in (CatalogType.OBJECT, CatalogType.SOURCE):
--> 181     catalog = _load_object_catalog(hc_catalog, config)
    182 elif catalog_type == CatalogType.MARGIN:
    183     catalog = _load_margin_catalog(hc_catalog, config)

File ~/projects/lincc-frameworks/lsdb/src/lsdb/loaders/hats/read_hats.py:268, in _load_object_catalog(hc_catalog, config)
    259         raise ValueError("The selected sky region has no coverage")
    260     hc_catalog = hc.catalog.Catalog(
[...]

PR #999, opened 2025-08-28
Add tests confirming OR filters, and NONE filters.

delucchi-cmu:
I was not able to reproduce behavior from either #776 or #739

Adds tests to confirm, and to prevent regressions.

### DIFF
--- a/tests/lsdb/loaders/hats/test_read_hats.py
+++ b/tests/lsdb/loaders/hats/test_read_hats.py
@@ -118,6 +118,74 @@ def test_read_hats_collection_with_extra_kwargs(small_sky_order1_collection_dir)
     assert np.all(catalog.margin.compute()["ra"] > 300)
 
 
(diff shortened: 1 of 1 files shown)


smcguire-cmu (commented):
I think the problems with filters don't really show up until we use HTTP paths, since most of them are to do with how we generate the HTML params for lsdb server. I think the None filters still is a problem in that case.

Also for HTTP filters for issue #776 we still don't allow [pyarrow Expressions](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Expression.html#) which are one of pyarrow's options for expressing filters. We could close out the issue though and just say we're not supporting it. From what I can tell there's not a very easy way of serializing it to use in lsdb server so it would be a lot of work for something probably no one would use.

smcguire-cmu (approved):
Looks good thanks!

PR #1003, opened 2025-09-03
Fix sphinx hiding output block in dask notebook

olivialynn:


olivialynn:
On refresh, our current version of the file works just fine 🙃  I have the original copy of the tab next to a refreshed/reloaded one, so I know I'm not crazy. 

Closing this, as the render issue is at most intermittent.

PR #1016, opened 2025-09-15
Use PEP-639 compatible license metadata

delucchi-cmu:
See also: https://github.com/lincc-frameworks/python-project-template/issues/532


drewoldag (approved):
Looks ok to me

PR #1020, opened 2025-09-17
Regression in windows pyarrow for chunked array

### LINKED ISSUE #974: LSDB CI pipeline failing for Windows+python3.10
**Bug report**

When preparing a PR for this project, the CI pipeline check for Windows with Python 3.10 fails, and the failure is in the test for the crossmatch algorithm, a discrepancy in the number of crossmatches (101 vs. 103). From `def test_kdtree_crossmatch_nested(algo, small_sky_catalog, small_sky_xmatch_catalog, xmatch_correct):`

```
>       assert np.sum(xmatched["small_sky_xmatch"].nest.list_lengths) == len(xmatch_correct)
E       assert np.int64(101) == 103
```

The other 3.10 builds do not fail. 

**Before submitting**
Please check the following:

delucchi-cmu:
The length of the array is 103, while the length of the index is 101. This is some kind of subtle regression in pyarrow v21.0.0 that I haven't been able to pinpoint exactly.

Closes #974

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
     "deprecated",
     "hats>=0.6.4",
     "nested-pandas>=0.4.7,<0.5.0",
-    "pyarrow>=14.0.1",
+    "pyarrow>=14.0.1,<21.0.0",
     "scipy>=1.7.2", # kdtree
     "universal-pathlib>=0.2.2",
 ]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
     "deprecated",
     "hats>=0.6.4",
     "nested-pandas>=0.4.7,<0.5.0",
-    "pyarrow>=14.0.1",
+    "pyarrow>=14.0.1,<21.0.0",
```
hombit, line 27: Do it for windows only?
```suggestion
    "pyarrow>=14.0.1,<21.0.0; platform_system == 'Windows'",
    "pyarrow>=14.0.1; platform_system != 'Windows'",
```

hombit (approved):
Should it close #974?

PR #1034, opened 2025-09-24
Carry loading config with catalog.

### LINKED ISSUE #736: Rename spatial objects from *Search to *Region
**Feature request**

"Region selection" terminology looks more intuitive for astronomers than "search filtering". I propose to change:

- [ ] `*Search -> *Region` or use [astropy regions](https://astropy-regions.readthedocs.io/en/stable/index.html) library classes instead
- [ ] `Catalog.read_hats(search_filter=) -> read_hats(region=)`
- [ ] `Catalog.*_search -> Catalog.*_region`
- [ ] `Catalog.search -> Catalog.select_region`

Related to https://github.com/astronomy-commons/lsdb/issues/664

**Before submitting**
Please check the following:

### LINKED ISSUE #1014: pyarrow filters not respected when applying `search`
**Bug report**

Reported by Phillip at CMU. He's doing something similar to this:

```
catalog = lsdb.open_catalog(..., columns=[...], filters=[<a lot of filters>])
catalog.search(cone_search_object)
```

For now, he's getting around it by adding `search_filter=cone_search_object`, but it was unexpected behavior.

This is likely because the pyarrow filters aren't carried through when we reload the catalog after the `search` invocation.

**Before submitting**
Please check the following:

delucchi-cmu:
Closes #1014 by carrying the `HatsLoadingConfig` used to load the catalog along with the rest of the member variables.

This change grew considerably because this introduced some circular dependencies. These circles were broken chiefly by the `*_Search` kinds of objects, since they were `__init__`'d in a few places. I did a small refactor to put all regions in one file, to ease imports elsewhere. This closes #736.

Much of the code in `read_hats.py` is simplified and de-duplicated, as another side-effect. Confirmation of #1014 occurs in `test_box_search.py`.


src/lsdb/loaders/hats/read_hats.py
```diff
@@ -50,6 +49,7 @@ def open_catalog(
     columns: list[str] | str | None = None,
     margin_cache: str | Path | UPath | None = None,
     error_empty_filter: bool = True,
+    filters=None,
```
smcguire-cmu, line 52: Can you add a docstring arg for this

tests/lsdb/catalog/test_box_search.py
```diff
@@ -52,6 +54,68 @@ def test_box_search_ra_complement(small_sky_order1_catalog):
...
+        filters=[("id", ">", 750)],
+        search_filter=lsdb.BoxSearch(ra=(0, 360), dec=(-80, -30)),
+    )
+    results = small_sky_order1_catalog.compute()
+    assert len(results) == 71
+    assert all(id > 750 for id in results["id"])
+    assert all(-80 <= ra <= -30 for ra in results["dec"])
+
+    # Confirm that, after adding another region search, the original filters are respected
+    ra_search_catalog = small_sky_order1_catalog.box_search(ra=(280, 300), dec=(-90, 90))
```
smcguire-cmu, line 75: I don't think this actually tests the reload on search since the catalog was loaded with a search_filter. Can you test this box_search call on the catalog that was loaded with only filters as well?

smcguire-cmu (commented):
I really like these changes, I think the code is much cleaner with more things in the loading config!

PR #1041, opened 2025-09-29
Update dask_client.ipynb

hombit:
Threads share memory, we cannot allocate memory per thread.


olivialynn:
Ah, my apologies. It was pending and not posted!

olivialynn (approved):

Issue #325, opened 2024-05-15
Refactor common code to hats library


smcguire-cmu:
There is a lot of code that performs tasks on the healpix level in LSDB that could be in hats. For example, code that performs coarse searching, filtering on margins, aligning partitions with margins and mocs for crossmatching/joining.

PR #1087, opened 2025-10-16
Move search functions from LSDB -> HATS

sedonaprice:
Moved lsdb.core.search submodule functions into a new hats.search submodule. The classes in the lsdb.core.search submodule remain in LSDB, as these are specifically implemented to work with LSDB catalogs.

Updated all relevant imports to import from hats.search and not lsdb.core.search.

This addresses part of https://github.com/astronomy-commons/lsdb/issues/325, and helps to meet stcal usage needs (enabling only HATS as a dependency).


delucchi-cmu (approved):
:rocket:

PR #1098, opened 2025-10-20
Read/write catalogs with nested default columns

### LINKED ISSUE #1029: Default columns when saving nested catalog
**Bug report**

When saving a nested catalog, we cannot currently set default nested columns:

```python
default_columns = ["ra", "dec", "sources.source_id", "sources.mjd"]
cat = small_sky_with_nested_sources[default_columns]
> cat.to_hats("small_sky", default_columns=default_columns)
-----------
ValueError: Default columns `{'sources.source_id', 'sources.mjd'}` not found in catalog
```

**Before submitting**
Please check the following:

camposandro:
Allow reading/writing catalogs with nested default columns. Closes #1029.

### DIFF
--- a/src/lsdb/nested/core.py
+++ b/src/lsdb/nested/core.py
@@ -437,8 +437,8 @@ Refer to the docstring for guidance on dtype requirements and assignment."""
 
     @property
     def exploded_columns(self) -> list[str]:
-        """returns the list of base column names and nested subcolumn names (in exploded, dot notation)"""
-        return self._meta.base_columns + self._meta.get_subcolumns()
+        """returns the list of column names and nested subcolumn names (in exploded, dot notation)"""
+        return self.columns.to_list() + self._meta.get_subcolumns()
 
     # NOTE: Naming conflict with LSDB.catalog.all_columns
     # Not wrapped in LSDB
--- a/tests/lsdb/nested/test_nestedframe.py
+++ b/tests/lsdb/nested/test_nestedframe.py
@@ -55,11 +55,12 @@ def test_nested_columns(test_dataset):
 
 
 def test_exploded_columns(test_dataset):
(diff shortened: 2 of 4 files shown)


src/lsdb/nested/core.py
```diff
@@ -435,6 +435,11 @@ def compute(self, **kwargs):
         """Compute this Dask collection, returning the underlying dataframe or series."""
         return npd.NestedFrame(super().compute(**kwargs))
 
+    @property
+    def exploded_columns(self) -> list[str]:
```
delucchi-cmu: I feel like I want a unit test for this method. My gut is that this would give `["id", "ra", "dec", "sources", "sources.mjd", "sources.mag"]`, but I'm not so sure about the `"sources"` entry.

tests/lsdb/nested/test_nestedframe.py
```diff
@@ -54,6 +54,16 @@ def test_nested_columns(test_dataset):
...
 
 
+def test_exploded_columns(test_dataset):
+    base_columns = test_dataset._meta.base_columns
+    assert base_columns == ["a", "b"]
+    subcolumns = test_dataset._meta.get_subcolumns()
+    assert test_dataset.nested_columns == ["nested"]
+    assert test_dataset["nested"].nest.columns == ["t", "flux", "band"]
+    assert subcolumns == ["nested.t", "nested.flux", "nested.band"]
+    assert test_dataset.exploded_columns == base_columns + subcolumns
```
delucchi-cmu: Ah! That's not the answer i want, then! When I talked to Kostya about this last wee, he said he would like to be able to specify "nested" as a default column, and that would pull in the entirety of column "nested". So maybe exploded should be `base_columns + nested_columns + subcolumns`?

PR #1137, opened 2025-11-14
Add a few instructions to release template

olivialynn:
- Adding some things that came up as I was going through the release process for the first time
- I'm just guessing on the procedure for the hash mismatch, as this was what seemed logical and worked for me--happy to change the explanation to something else
- Also happy to change any of the other wording here


delucchi-cmu (approved):

delucchi-cmu (approved):
Thanks! Looks great!

Issue #1083, opened 2025-10-16
Unevaluated (lazy) view of Catalog should show upper size limit as well as partitions


gitosaurus:
**Feature request**

When preparing to pull data from a catalog, especially a remotely located one, it would be very good to have some estimate of how much data will be transferred.  While an exact estimate is not possible without actually performing the computation (because of the effect of row filtering, for example) the effects of region searches and column selection can be well-estimated, using the metadata from the selected partitions.

When a lazy representation of a Catalog is shown in a notebook, the number of partitions is reported, so the partition collection is known.  Each of these partitions has metadata describing the data consumed for each column.  Therefore it is possible to report an *upper limit* on the data that will be required for a given region+column selection.  While not the same as complete preflighting, this upper bound would be valuable to users, who might choose to reduce their region or column selection further before starting computation.

Note that this information does not appear obtainable today via the `Catalog` attributes; it is something that will need to come from looking at the metadata in the Parquet files underlying the partitions.

From verbal discussions during LSDB Busy Week 2025 at STSci.

**Before submitting**
Please check the following:

- [X] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [X] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [X] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

gitosaurus:
The following code is one implementation of what is desired:
```python
def get_catalog_metadata(cat_orig):
    return pq.read_metadata(cat_orig.hc_structure.catalog_path / "dataset" / "_metadata")
    
def total_io_size(rg, column_nums: set[int]):
    return sum([rg.column(c).total_compressed_size for c in column_nums])

def size_estimate(cat_orig, md, cat):
    """Gives an upper-bound estimate in bytes of the size of the catalog
    with its partition and column selections.

    cat_orig: the original catalog on which `cat` is based.
    md: the result of get_catalog_metadata(cat_orig)
    cat: the catalog based on cat_orig which has been scoped by search_filter=
         and/or columns=.
    """
    pixel_list = [cat_orig.get_partition_index(*(p.get_healpix_pixels()[0])) for p in cat.partitions]
    include_columns = set(cat.columns)
    col_nums = {ic for ic, cname in enumerate(md.schema.names) if cname in include_columns}
[...]

Issue #389, opened 2024-07-30
Catalog should provide an approximate size, even after filtering applied


hombit:
**Feature request**

It would be really nice if `Catalog` would provide a range of possible values for:

- [ ] Number or rows
- [ ] Bytes in memory after `.to_compute()`
- [ ] Bytes on disk after `.to_hipscat()`
- [ ] Bytes to (down)load from network/disk

I understand that we cannot know how many rows we would have after filtering by values, with `.query()` or `read_hipscat(filters=...)`. However we do know how many "full" partitions would left after geometrical filtering, see e.g. [`cdshealpix.nested.cone_search`](https://cds-astro.github.io/cds-healpix-python/stubs/cdshealpix.nested.cone_search.html#cdshealpix.nested.cone_search) docs.

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [x] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [ ] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

PR #1179, opened 2025-12-17
Estimate catalog size

camposandro:
This PR adds a new `estimate_size` method to `HealpixDataset` that provides a rough estimate of the number of rows, memory size, and disk size of an in-memory catalog. The main goal is to give users a sense of the order of magnitude of data reduction when using spatial filters and column selection.

The results are holistic and based on the "_metadata" statistics of the raw parquet data pages. Therefore, we cannot provide a full memory footprint, which may also vary between machines and readers.

We do not print the size estimates directly on the Catalog lazy view (as initially suggested in #1083) since querying the metadata is a very slow operation. Instead, this utility is available as a separate method.

This is a first step toward resolving issues #389 and #1083.

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -1523,7 +1523,7 @@ class HealpixDataset:
         # Row count estimate
         orig_cat = hc.read_hats(self.hc_structure.catalog_path)
         expected_cat_rows = int(get_row_count(pixel_stats))
-        row_ratio = expected_cat_rows / int(orig_cat.catalog_info.total_rows) * 100
+        row_pct = expected_cat_rows / int(orig_cat.catalog_info.total_rows) * 100
         expected_cat_rows = int_comma(expected_cat_rows)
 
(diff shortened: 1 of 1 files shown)


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -1480,3 +1481,59 @@ def plot_points(
...
+                use_default_columns=False,
+                exclude_hats_columns=False,
+                include_columns=self.columns,
+                include_stats=("row_count", "disk_bytes", "memory_bytes"),
+            )
+
+        # Row count estimate
+        orig_cat = hc.read_hats(self.hc_structure.catalog_path)
+        expected_cat_rows = int(get_row_count(pixel_stats))
+        row_ratio = expected_cat_rows / int(orig_cat.catalog_info.total_rows) * 100
```
gitosaurus:
```suggestion
        row_pct = expected_cat_rows / int(orig_cat.catalog_info.total_rows) * 100
```
Or another suffix that indicates percentage rather than ratio.

PR #1184, opened 2025-12-23
Update ra and dec column in catalog_info during rename

### LINKED ISSUE #1026: `Catalog.rename` with ra and dec columns doesn't update metadata
When calling `Catalog.rename` with the ra and dec columns, the names of the ra and dec columns in the TableProperties metadata is not updated. We should either update them if they are renamed or not allow renaming these columns.

smcguire-cmu:
Closes #1026


gitosaurus (approved):
LGTM

Issue #463, opened 2024-10-25
Extend and document join behavior better


hombit:
`Catalog.join` has few usability/documentation issues:

- [ ] Docstring doesn't say that the join type is inner
- [ ] Left join would be useful. Example: join Gaia source catalog (~2B rows) and [Gaia Cepheids] https://github.com/astronomy-commons/lsdb/issues/655(http://vizier.cds.unistra.fr/viz-bin/VizieR-3?-source=I/358/vcep) (15k rows)
- [ ] Outer join could be useful and may also solve #132. Example: join [Gaia Cepheids](http://vizier.cds.unistra.fr/viz-bin/VizieR-3?-source=I/358/vcep) and [Gaia RR Lyr](http://vizier.cds.unistra.fr/viz-bin/VizieR-3?-source=I/358/vrrlyr)
- [ ] https://github.com/astronomy-commons/lsdb/issues/785
- [ ] Right join?

This is originated from LINCC Catalyst / DSFP hack day

nevencaplar:
Closing as we completed all of the subtickets (right join is not implemented)

PR #1208, opened 2026-01-19
Mention inner join in catalog.join docstring

nevencaplar:
Closes first bulletpoint in https://github.com/astronomy-commons/lsdb/issues/463


hombit (approved):
Thanks

PR #1210, opened 2026-01-19
Update map_partitions nb to use collection

### LINKED ISSUE #1117: Update pre-ex nb: map_partitions
On testing this notebook as part of #1114, I found:
  - FileNotFoundError: Failed to read HATS at location `/data3/epyc/data3/hats/catalogs/gaia_dr3`
  - Which, nbd, we can update the catalog_root to the commented-out alternative on the line just above
  - But then I found I kept getting a `FSTimeoutError` later down in the notebook (the `timeit`'d cell where we close the client, then reopen and run `result = unrealized.compute()` to search the whole catalog) - guessing that's due to no longer simply running things on the same file system.

I think my judgement is a little fatigued from looking at documentation for as long as I have. Potentially we could just remove this part? Or readjust the parameters we're feeding to the client?

nevencaplar:
Changed link to use the collection. Reran the notebook, verified that it works. Added few explanations. 

Closes https://github.com/astronomy-commons/lsdb/issues/1117.

### DIFF
--- a/docs/tutorials/pre_executed/map_partitions.ipynb
+++ b/docs/tutorials/pre_executed/map_partitions.ipynb
@@ -52,7 +52,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 4,
    "id": "ff803be2-c4c9-4bf8-8fe0-5b7e896d6478",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


olivialynn (approved):
Fantastic, this looks great! Thanks!

PR #1211, opened 2026-01-20
Check (and minor fixes to) pre-executed notebooks that require SDF

olivialynn:
# Overview
- These were **run on data.lsdb.cloud** on a Large allocation (4.0 CPU, 16Gi RAM) 
  - (As of such, I did not re-run the non-RSP sections in `rubin_dp1.ipynb`)
- **Legacy API Catalog.dropna has been removed** from:
  - `using_rubin_data.ipynb` 
  - (note: neither use of `dropna` was removed in `visualize_periodic_lcs.ipynb`, as that turned out to be non-Catalog calls)
- **Dask parameters were re-tuned** in the following notebooks, so they could be run to completion on a Large allocation on the RSP (which, of course, is not so large...):
  - `rubin_dp1_vsx.ipynb`
  - `visualize_periodic_lcs.ipynb`
- **"Author" and "last run" sections were added** where needed, and **"last run" sections were updated** to current dates
- **Catalog paths were updated** to match slight but breaking changes in our naming schemes:
  - `vsx_2025-03-21/vsx/` → `vsx` in `rubin_dp1_vsx.ipynb`
  - `object_photoz_no_mags` → `object_photoz` in `rubin_dp1_photoz.ipynb`

### Dask parameters for rubin_dp1_vsx
I tested:
- workers=8, memory_limit=auto
- workers=4, memory_limit=auto 
- **workers=4, memory_limit=4GB (what I picked, and shown below)** 
- workers=8, memory_limit=2GB (workers started getting killed/restarted here) 

The first 3 options seemed pretty similar, with 4 workers/4GB memory_limit being the fastest, so I went with that—though, it did still give more of a red-heavy task stream than might be preferred (?)
Screenshot 2026-01-07 at 10 40 21 AM

### Dask parameters for visualize_periodic_lcs
**I reduced the workers from 20 to 4** so the notebook could run on a standard large partition (4.0 CPU, 16Gi RAM) , but this does create some cluttered output when we first start the client. 

Maybe this could be fine-tuned away? Not sure if that's possible; not sure if that's worth the time—but happy to look into it if someone thinks it is.


nevencaplar:
That is good. Whatever it takes to remove superfluous DASK output. You didnt mention doing the same in `Cross-matching with Variable Star Index catalog`, just making sure that it is removed in all notebooks.

delucchi-cmu (approved):

PR #1213, opened 2026-01-21
Fix landing page

nevencaplar:
One of my previous pull has modified the landing page for documentation. This reverts to original state. 

Previous state:
image_2026-01-20_165631268

After state:
Screenshot 2026-01-20 at 4 57 37 PM


wilsonbb (approved):
This looks good to me

There are pre-commit failures from mypy but they seem un-related to this PR

Issue #1222, opened 2026-01-23
Use PyPI source for conda-forge receipts


hombit:
Pretty often we have a problem with SHA256 mismatch when releasing conda packages, example:
https://github.com/conda-forge/hats-feedstock/pull/61/changes/ff3f45eb171a020d819aa665494c7642b02f3cf5

We should switch from github to pypi as a package source, this will fix the issue. See example in lightcurvelynx repo:
https://github.com/conda-forge/lightcurvelynx-feedstock/blob/main/recipe/meta.yaml

PR #1231, opened 2026-01-27
Copy small_sky_order1_default_columns/dataset and regenerate metadata

### LINKED ISSUE #1230: Some test data is missed from .tar.gz on PyPI
**Bug report**

Switching conda-forge LSDB receipt to use PyPI source fails because symlinked test-data is missed from the PyPI .tar.gz distribution:
https://github.com/conda-forge/lsdb-feedstock/pull/81#issuecomment-3807751050

Also symlinks don't properly work on Windows.

We should copy `tests/data/small_sky_order1_default_columns/dataset` folder instead of symlinking it.

**Environment Information**

<details>
<summary>Traceback</summary>

FILL IN YOUR STACK TRACE HERE

</details>

**Before submitting**
Please check the following:

- [ ] I have included information about my environment, including the version of this package (e.g. `lsdb.__version__`)

hombit:
It copies the dataset instead of linking it, so the data finds its way to the .tar.gz file posted on PyPI. I haven't regenerated all the test data; let me know if I should.

Closes #1230 and enables https://github.com/conda-forge/lsdb-feedstock/pull/81 and #1222


camposandro (approved):
Looks good!

PR #1246, opened 2026-02-05
Add CatalogStream Interface

### LINKED ISSUE #1042: Add a background running data iterator interface
**Feature request**

LSDB would be a great fit for machine learning training pipelines as a streaming dataset. I propose to implement a streaming interface as a Python iterator object, which would fetch data from a few random partitions in the background and yield the result. We can implement both an infinite iterator and an iterator that goes through the catalog exactly once.  

We can also provide integration with PyTorch as a Catalog method returning a torch data loader object.  

See the prototype of this idea in the Uncle Val repo:
https://github.com/lincc-frameworks/uncle-val/blob/main/src/uncle_val/learning/lsdb_dataset.py

**Before submitting**
Please check the following:

dougbrn:
Closes #1042. Generally pretty open on a lot of the design decisions, naming, etc here. I didn't attempt the pytorch integration mentioned in the issue, happy to discuss more about that!


hombit (commented):
I have a few overall design questions
 
1. Maybe we should implement an iterator and iterable separately, so the iterable object may be reused.  
2. I'm not sure if there is a good use case for non-None `iter_limit > 1`: we could get row duplicates with it even in a single batch, which is not the case for `1` and `None`. I believe that for most ML approaches `iter_limit = 1` and `iter_limit = None` are enough, and if people would like to do multiple epochs, they would reuse an iterable object n times.  
3. I understand that it may be out of scope and looks more like a Dask feature request, but I think we should also support anything a user can get from a Catalog object, including a Dask series they can get with `['column']` or `.map_partitions`.

dougbrn:
Thanks for taking a look!

1. I agree with this, not 100% sure on where the dividing line between the two will be and how the re-use will look API wise but I can make an attempt at it.
2. Fair point, I had originally introduced this as a way to allow larger iteration without stepping on the API rake of this way of interacting with a single iteration:
```
cat_iter = lsdb.CatalogIterator(cat, loop=False)

for chunk in cat_iter:
     # do something
```
resulting in an endless loop after simply switching the kwarg value
```
cat_iter = lsdb.CatalogIterator(cat, loop=True)

for chunk in cat_iter:
     # do something
```
but it changing the behavior makes it probably not worth it and users will just need to be aware of what they're doing by switching that bool.
3. I'm not sold on this. I do understand that there will be situations where users may have a series as a result of catalog work, but it feels like blurring the lines between what a user can do in LSDB with a series vs with a catalog. For example, even if something like `catalog["nested"].crossmatch` might feel potentially right to do, we don't allow crossmatch to flex for a `dd.series` (I know it's not a perfect comparison). It's very trivial in column selection and in map_partitions to return a dataframe (catalog) in place of a series, from my perspective, as well.

hombit:
It also may be `infinite` instead of `loop`. Or we can just drop that functionality and focus on a one-time catalog scanning

hombit (commented):
I do like this implementation!

I think that the Stream object should not be changed by an Iterator object. I propose to "split" `rng` when creating an iterator and pass a new one inside it, and then pass it back to `get_next_partitions`. So basically, the stream's `rng` is used to initialize new iterators, and the iterator's `rng` is used for all the shuffling.

dougbrn:
@hombit Now the rng should be split

hombit:
@dougbrn `rng.spawn(1)[0]`

src/lsdb/streams/catalog_streams.py
```diff
@@ -0,0 +1,242 @@
...
+        if len(self.partitions_left) > 0:
+            self.future = self.iterable.submit_next_partitions(self._get_next_partitions())
+        else:
+            self._empty = True
+            self.future = None
+
+        return result
+
+    def __len__(self) -> int:
+        return int(np.ceil(len(self.partitions_left) / self.iterable.partitions_per_chunk))
```
hombit: Should it fail on NotImpleneted for the infinite iterator?

hombit (approved):
Looks good, thank you!

PR #1254, opened 2026-02-17
Fix ztf catalog column names used in getting started

skoranda:
Fix the ztf catalog column names used in the Performing Filters section of Getting Started so that a simple cut and paste from documentation into a Python session does not result in stack trace.


nevencaplar (approved):
Good catch. Thank you!

PR #1276, opened 2026-03-05
Add resume functionality to `to_hats`

### LINKED ISSUE #760: Make it possible to resume LSDB pipeline
**Feature request**

Let's say I'm running a pipeline, but the job crashes. It would be nice to resume and continue from the same point!

One option would be to add a `Catalog.to_hats(..., resume: bool = False)` argument which, if set to `True`, would check the target folder and skip all partitions that are already present. It should probably also verify the validity of existing Parquet files.

**Before submitting**
Please check the following:

smcguire-cmu:
Changes to_hats to write done and histogram files while writing, allowing the pipeline to be resumed by reading the progress in the directory.

Closes #760


delucchi-cmu (commented):
Please tag the related issue. Looks like #760.

delucchi-cmu (approved):

PR #1284, opened 2026-03-11
Introduce crossmatch_nested(how: "left" | "inner")

### LINKED ISSUE #785: `crossmatch_nested` mixes "inner" and "left" join behaviors
**Bug report**

Currently `Catalog.crossmatch_nested` makes an "inner" join on the HEALPix tree level (1) and "left" join on the partition level (2). It is a confusing behavior.

1. 
https://github.com/astronomy-commons/lsdb/blob/ba4dbe6e017633d52b5369911a0da2cf8733e64b/src/lsdb/dask/crossmatch_catalog_data.py#L215
https://github.com/astronomy-commons/lsdb/blob/ba4dbe6e017633d52b5369911a0da2cf8733e64b/src/lsdb/dask/merge_catalog_functions.py#L106

2.
https://github.com/astronomy-commons/lsdb/blob/ba4dbe6e017633d52b5369911a0da2cf8733e64b/src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py#L255

Related to https://github.com/astronomy-commons/lsdb/issues/463 and https://github.com/astronomy-commons/lsdb/issues/655

**Before submitting**
Please check the following:

### LINKED ISSUE #1149: Add `how='inner'` argument to crossmatch_nested
**Feature request**

Following the fix to #655, add the same `how=` argument to `crossmatch_nested`.

Per @camposandro: nested crossmatch is already a left-join but only for the pixels that result from the inner alignment of the two catalogs. Does this mean that if we have pixels of the left catalog that do not overlap with those on the right then the data of those left pixels would be missing in the results of a "left crossmatch"? We should determine that behavior.

See https://github.com/astronomy-commons/lsdb/pull/1138#discussion_r2542771996

**Before submitting**
Please check the following:

- [X] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

hombit:
Fixes https://github.com/astronomy-commons/lsdb/issues/785, it is a breaking change, because `crossmatched_nested` results will change from what we have now: no `Nulls` for `how='inner'` (default) and adds non-overlapping pixels for `how='left'`.

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -399,6 +399,7 @@ class Catalog(HealpixDataset):
         output_catalog_name: str | None = None,
         require_right_margin: bool = False,
         nested_column_name: str | None = None,
+        how: str = "inner",
     ) -> Catalog:
         # pylint:disable=unused-argument
         """Perform a cross-match between two catalogs, adding the result as a nested column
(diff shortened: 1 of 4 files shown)


smcguire-cmu (approved):
Looks good thanks!

PR #1302, opened 2026-03-16
fix: handle suffix collision in catalog self-crossmatches

### LINKED ISSUE #1275: Catalog self crossmatches fail due to suffix defaults
**Bug report**
The following simple example fails with `ValueError: cannot reindex on an axis with duplicate labels`:
```
import lsdb
cat = lsdb.generate_catalog(5000,100, lowest_order=4, ra_range=(15.0,25.0), dec_range=(34.0,44.0), seed=1)
cat.crossmatch(cat).compute()
```
This is due to the default suffix behavior causing collisions, so the suffixes need to be manually provided:
`cat.crossmatch(cat, suffixes=("_l", "_r")).compute()`

This should work by default, ideally. We could check for suffix collision and append (_l, _r) to them, for example

**Before submitting**
Please check the following:

- [ ] I have included information about my environment, including the version of this package (e.g. `lsdb.show_versions()`)

travellingsoldier85:
## Summary

Fixes #1275 — Catalog self-crossmatches fail with `ValueError: cannot reindex on an axis with duplicate labels` when using default suffixes.

## Problem

When a catalog is crossmatched against itself:
```python
cat = lsdb.generate_catalog(...)
cat.crossmatch(cat).compute()  # ValueError!
```

`_default_suffixes()` generates identical suffixes (both `_<name>`), which causes duplicate column labels and a `ValueError` in pandas.

## Solution

When `_default_suffixes()` detects that both catalog names are the same, it appends `_l` and `_r` to produce unique suffixes:
- Before: `("_small_sky", "_small_sky")` → collision
- After: `("_small_sky_l", "_small_sky_r")` → works

When names differ, the existing behavior is preserved unchanged.

## Changes

- **`src/lsdb/catalog/catalog.py`**: Updated `_default_suffixes()` to detect and resolve suffix collisions
- **`tests/lsdb/catalog/test_crossmatch.py`**: Added tests for both self-crossmatch and normal cases

## Testing

Added two test functions:
- `test_self_crossmatch_default_suffixes` — verifies unique suffixes are generated for self-crossmatch
- `test_default_suffixes_different_names` — verifies existing behavior is preserved for different catalog names

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -43,8 +43,20 @@ from lsdb.types import DaskDFPixelMap
 
 
 def _default_suffixes(left_name: str, right_name: str) -> tuple[str, str]:
-    """Return the default pair of suffixes for left/right catalog names."""
-    return (f"_{left_name}", f"_{right_name}")
+    """Return the default pair of suffixes for left/right catalog names.
+
+    When both names are identical (e.g. a self-crossmatch), ``_l`` and
+    ``_r`` are appended to avoid suffix collisions that would cause
+    ``ValueError: cannot reindex on an axis with duplicate labels``.
+
+    See https://github.com/astronomy-commons/lsdb/issues/1275
+    """
+    left_suffix = f"_{left_name}"
+    right_suffix = f"_{right_name}"
+    if left_suffix == right_suffix:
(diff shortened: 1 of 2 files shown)


hombit:
Closing it for now. Feel free to re-open or re-submit with the fixes requested above.

PR #1303, opened 2026-03-17
Refactor release tracker template for clarity

delucchi-cmu:
Updated the release tracker template to streamline the release process.

All conda-forge releases now happen in a single feedstock.


olivialynn (approved):
Yay! Also, I like the handy links for github releases!

PR #1308, opened 2026-03-18
Small improvements to epoch propagation notebook

camposandro:
Some minor updates to the DP1 x Gaia epoch propagation notebook:

- Remove some unused imports
- Be more explicit about pre-selection of columns in `open_catalog`
- Add important note to the user to not change coordinates in place
- Add axes labels to plot of distance distribution
- Add comparison of median values


hombit (approved):
Thank you!

PR #1325, opened 2026-04-06
Add Kubernetes deployment guide for Dask Operator

mithr4ndir:
Closes #1324

## Summary
Adds a practical guide for deploying lsdb workloads on Kubernetes using the Dask Kubernetes Operator. Covers building a custom container image, installing the operator via Helm, creating a DaskCluster with autoscaling, connecting lsdb to the cluster, resource sizing recommendations, monitoring, and common troubleshooting scenarios.

Positioned as a complement to the existing `dask-cluster-tips.rst` (local and SLURM setups), with appropriate cross-references.

Based on a working deployment in my home lab — 3-node K8s cluster with the Dask Operator, validated with spatial queries, crossmatches, and catalog operations.

## Changes
- New file: `docs/tutorials/kubernetes-deployment.rst`
- Updated: `docs/tutorial_toc/toc_performance.rst` (added to Performance Tips TOC)

## Test plan
- [x] Pre-commit checks pass
- [x] Sphinx doc build succeeds without new warnings
- [ ] Links and cross-references render correctly on ReadTheDocs

### DIFF
--- a/docs/tutorials/kubernetes-deployment.rst
+++ b/docs/tutorials/kubernetes-deployment.rst
@@ -1,0 +1,309 @@
+Kubernetes deployment with the Dask Operator
+============================================
+
+For workloads that exceed what a single machine or a traditional HPC cluster can offer,
+Kubernetes provides a flexible platform for running LSDB at scale.
+A Kubernetes cluster can dynamically allocate resources, scale workers up and down based on demand,
+and provide a reproducible environment for data processing pipelines.
+
+This guide walks through deploying a `Dask Kubernetes Operator <https://kubernetes.dask.org/en/latest/operator.html>`__
+cluster suitable for LSDB workloads.
+It complements the :doc:`Dask cluster configuration tips </tutorials/dask-cluster-tips>` page,
+which covers local and SLURM-based setups.
+
+.. note::
+
+    Kubernetes adds operational complexity compared to a local cluster or SLURM job.
+    If your data fits in memory on a single machine, a
+    `LocalCluster <https://distributed.dask.org/en/stable/api.html#distributed.LocalCluster>`__
+    is simpler and often sufficient.
+    Consider Kubernetes when you need multi-node scaling, autoscaling, or a shared
+    platform that multiple users or pipelines can share.
+
+Prerequisites
+-------------
+
+Before starting, you will need:
+
+- A running Kubernetes cluster (v1.25+) with ``kubectl`` configured
+- `Helm <https://helm.sh/docs/intro/install/>`__ v3 installed
+- A container registry you can push images to (Docker Hub, GitHub Container Registry, etc.)
+- Basic familiarity with Kubernetes concepts (pods, services, persistent volumes)
+
+Building a container image
+--------------------------
+
+The Dask project does not ship an official image with LSDB pre-installed.
+Instead, you build your own image with the packages your workflow requires.
+This keeps the image minimal and lets you pin exact versions.
+
+A minimal ``Dockerfile``:
+
+.. code-block:: dockerfile
+
+    FROM python:3.12-slim
+
+    RUN pip install --no-cache-dir \
+        lsdb \
+        dask[complete] \
+        distributed
+
+    # Add any additional packages your workflow needs, for example:
+    # RUN pip install --no-cache-dir s3fs jupyter
+
+Build and push the image to your registry:
+
+.. code-block:: bash
+
+    docker build -t registry.example.com/lsdb-dask:latest .
+    docker push registry.example.com/lsdb-dask:latest
+
+.. tip::
+
+    Pin package versions in production images (e.g. ``lsdb==0.5.0``) so that
+    your results are reproducible across runs.
+
+Installing the Dask Kubernetes Operator
+---------------------------------------
+
+The `Dask Kubernetes Operator <https://kubernetes.dask.org/en/latest/operator.html>`__
+manages ``DaskCluster`` custom resources, handling scheduler and worker pod lifecycles
+and optional autoscaling.
+
+Install the operator with Helm:
+
+.. code-block:: bash
+
+    helm repo add dask https://helm.dask.org
+    helm repo update
+
+    helm install dask-operator dask/dask-kubernetes-operator \
+        --namespace dask-operator \
+        --create-namespace
+
+Verify that the operator pod is running:
+
+.. code-block:: bash
+
+    kubectl get pods -n dask-operator
+
+Creating a DaskCluster
+----------------------
+
+The following manifest creates a Dask cluster with a scheduler, workers, and an autoscaler.
+Adjust resource requests, replica counts, and the image reference to match your environment.
+
+.. code-block:: yaml
+
+    apiVersion: kubernetes.dask.org/v1
+    kind: DaskCluster
+    metadata:
+      name: lsdb-cluster
+    spec:
+      worker:
+        replicas: 2
+        spec:
+          containers:
+            - name: worker
+              image: "registry.example.com/lsdb-dask:latest"
+              args:
+                - dask-worker
+                - --name
+                - $(DASK_WORKER_NAME)
+                - --nthreads
+                - "2"
+              resources:
+                requests:
+                  cpu: "2"
+                  memory: "4Gi"
+                limits:
+                  memory: "4Gi"
+              volumeMounts:
+                - name: catalog-data
+                  mountPath: /data
+          volumes:
+            - name: catalog-data
+              persistentVolumeClaim:
+                claimName: hats-catalog-pvc
+      scheduler:
+        spec:
+          containers:
+            - name: scheduler
+              image: "registry.example.com/lsdb-dask:latest"
+              args:
+                - dask-scheduler
+              resources:
+                requests:
+                  cpu: "2"
+                  memory: "6Gi"
+                limits:
+                  memory: "6Gi"
+              ports:
+                - name: tcp-comm
+                  containerPort: 8786
+                  protocol: TCP
+                - name: http-dashboard
+                  containerPort: 8787
+                  protocol: TCP
+        service:
+          type: ClusterIP
+          selector:
+            dask.org/cluster-name: lsdb-cluster
+            dask.org/component: scheduler
+          ports:
+            - name: tcp-comm
+              protocol: TCP
+              port: 8786
+              targetPort: tcp-comm
+            - name: http-dashboard
+              protocol: TCP
+              port: 8787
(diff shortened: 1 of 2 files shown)


docs/tutorials/kubernetes-deployment.rst
```diff
@@ -0,0 +1,309 @@
...
+- Reduce ``--nthreads`` to lower per-worker memory pressure.
+
+Tasks are slow or idle workers
+..............................
+
+If you see many idle workers while tasks queue up, the graph may not have enough partitions
+to keep all workers busy.
+
+- Check ``catalog.npartitions`` to see how many partitions are available.
+- Use ``lsdb.open_catalog(search_filter=...)`` to narrow the spatial region before
```
olivialynn: The suggested fix of narrowing the spatial region with `search_filter` doesn't directly address the idle worker symptom described at the top of this section. If too few partitions exist to keep all workers busy, reducing the spatial region would decrease the partition count further, potentially making the problem worse. The `search_filter` suggestion may be better suited to a separate tip about slow tasks or memory pressure, while the idle worker diagnosis focuses specifically on `catalog.npartitions` relative to worker count.

mithr4ndir:
Just pushed some fixes based on end-to-end testing against a live Gaia DR3 catalog:

- The Python example was causing OOM on workers when opening the full catalog remotely. Updated to use `search_filter` at `open_catalog` time with column projection, which is the recommended pattern for large remote catalogs. I did not catch this initially because my first round of testing used small synthetic catalogs on local storage, not Gaia over HTTP.
- Added a note that the Dask operator is cluster-scoped (only needs one install per cluster).
- Added a tip about `imagePullSecrets` for private container registries.
- Aligned the autoscaler minimum with the initial worker replica count.

Let me know if you run into anything else on your cluster.

docs/tutorials/kubernetes-deployment.rst
```diff
@@ -0,0 +1,327 @@
...
+
+    If your container image is in a private registry, add ``imagePullSecrets``
+    to both the scheduler and worker pod specs. Public images (e.g. on Docker Hub
+    or GitHub Container Registry with public visibility) do not need this.
+
+The PersistentVolumeClaim ``hats-catalog-pvc`` should point to storage containing your
+HATS catalogs. If your catalogs are accessed over the network (e.g. via
+``https://data.lsdb.io``), you can remove the volume mount entirely.
+
+Connecting LSDB to the cluster
```
olivialynn, line 300: Since these are mutually exclusive, it would be nice to label them explicitly as distinct options (e.g. "Option A: From inside the cluster" and "Option B: From outside the cluster"). This would help prevent readers who are skimming (or just less experienced with Kubernetes) from accidentally mixing the connection string from one approach with the setup steps from the other.

docs/tutorials/kubernetes-deployment.rst
```diff
@@ -0,0 +1,327 @@
...
+    client = Client("tcp://lsdb-cluster-scheduler:8786")
+
+    # Use search_filter at open time to avoid loading the full catalog into memory.
+    # Adding columns= further reduces the memory footprint per partition.
+    catalog = lsdb.open_catalog(
+        "/data/catalogs/my_catalog",
+        search_filter=lsdb.ConeSearch(ra=180, dec=0, radius_arcsec=600),
+        columns=["ra", "dec", "phot_g_mean_mag"],
+    )
+    result = catalog.compute()
```
olivialynn, line 328: Since this snippet is intended as a verification step, adding a line to print the first couple rows would be helpful to give users immediate, human-readable confirmation that the cluster is functioning as expected.

docs/tutorials/kubernetes-deployment.rst
```diff
@@ -0,0 +1,327 @@
...
+-------------
+
+Before starting, you will need:
+
+- A running Kubernetes cluster (v1.25+) with ``kubectl`` configured
+- `Helm <https://helm.sh/docs/intro/install/>`__ v3 installed
+- A container registry you can push images to (Docker Hub, GitHub Container Registry, etc.)
+- Basic familiarity with Kubernetes concepts (pods, services, persistent volumes)
+
+Building a container image
```
olivialynn, line 33: The troubleshooting section mentions that version mismatches between the client environment and the cluster can cause serialization errors, but this is easy to encounter and worth addressing proactively. It may be worth adding a note somewhere in this section recommending that users also pin the Python base image version (e.g. is it possible to do `FROM python:3.12.3-slim` rather than `FROM python:3.12-slim`?) to match their local environment. This would prevent the `VersionMismatchWarning` that users are likely to encounter during their first connection attempt.

docs/tutorials/kubernetes-deployment.rst
```diff
@@ -0,0 +1,327 @@
...
+                  cpu: "2"
+                  memory: "4Gi"
+                limits:
+                  memory: "4Gi"
+              volumeMounts:
+                - name: catalog-data
+                  mountPath: /data
+          volumes:
+            - name: catalog-data
+              persistentVolumeClaim:
```
olivialynn, line 141: We reference this here in the manifest, but the document doesn't include guidance on creating the corresponding PersistentVolume and PersistentVolumeClaim. While experienced Kubernetes users may be comfortable writing this from scratch, the stated prerequisite is only "basic familiarity with Kubernetes concepts," and in practice, most users prefer to adapt a concrete example. Would it be worth adding a small section (or at least an example manifest) covering a representative PV/PVC setup?

olivialynn (commented):
Thanks! This looks good and worked for me. I have a few comments on things that stuck out to me as I went through the steps described; interested to hear your thoughts!

nevencaplar:
@mithr4ndir Hi Chris! Do you think you would still be able to have a look at the comments that Olivia made above?

mithr4ndir:
> @mithr4ndir Hi Chris! Do you think you would still be able to have a look at the comments that Olivia made above?

Absolutely @nevencaplar , apologies for the delay @olivialynn , allow me a few hours and I'll respond to each comment.

docs/tutorials/kubernetes-deployment.rst
mithr4ndir: Done. The two paths are now explicit subsections, "Option A: From inside the cluster" and "Option B: From outside the cluster", each with its own connection string example, and a lead in sentence that calls out that they are mutually exclusive. Option B references back to Option A so the rest of the LSDB code does not have to be repeated.

docs/tutorials/kubernetes-deployment.rst
mithr4ndir: Added print(result.head()) to the verification snippet, plus one sentence explaining what it confirms (catalog open + distributed compute + result collection). Smoke tested against a real cluster: a 600 arcsec cone around M67 returns 1103 stars from public Gaia DR3 data and head() prints a clean five-row preview. Also confirmed it handles the empty-cone edge cleanly (prints Empty NestedFrame / Columns / Index rather than crashing).

docs/tutorials/kubernetes-deployment.rst
mithr4ndir: Good catch, thanks. Updated the Dockerfile example to python:3.12.3-slim and added a tip beneath it explaining why pinning the patch release matters for VersionMismatchWarning and serialization errors. I verified end-to-end against my homelab: connecting from a Python 3.12.3 client to a worker built on Python 3.12.12 emits the warning every time; rebuilding the client to match (3.12.3 ↔ 3.12.3) makes it disappear

docs/tutorials/kubernetes-deployment.rst
mithr4ndir: Agreed, basic familiarity is a reasonable bar but a copy-pasteable storage example is much friendlier. Added a new "Providing the catalog volume" subsection that covers two cases: a simple RWX PVC against a default StorageClass, and a static NFS PV+PVC pair for clusters without dynamic shared-storage provisioning. Also added a note steering object-storage users away from PVCs entirely. Validated both manifests against my cluster (NFS-backed, RWX, on Kubernetes 1.33). Also added a tip about the common "PVC stuck in Pending because there's no default StorageClass" failure mode, which is what I hit on first apply.

docs/tutorials/kubernetes-deployment.rst
mithr4ndir: That makes sense, the search_filter advice was in the wrong section. Split the troubleshooting block into "Idle workers" (focused on catalog.npartitions versus worker thread count) and a new "Slow tasks or memory pressure" block where search_filter and columns= actually belong. Also added an explicit note in the idle workers block that search_filter would reduce partition count further, in case anyone is tempted to try it. While there I tightened the wording on partitioning: HATS catalog partition count is fixed at import time by HEALPix order, so the recommendation is now "scale the cluster down" or "re-import the source with a finer HEALPix order via hats-import", rather than implying a runtime repartition that does not exist.

mithr4ndir:
Thanks @olivialynn for the careful review and @nevencaplar for the nudge, apologies again for the delay. New commit pushed addressing all five threads, and the branch was rebased onto current main while I was at it (head is now 3b011408):

 - Pinned python:3.12.3-slim in the Dockerfile and added a tip explaining why
 - Added a "Providing the catalog volume" subsection with concrete dynamic-RWX and static-NFS PV/PVC manifests, plus an object-storage note
 - Labelled the two connection paths "Option A: From inside the cluster" and "Option B: From outside the cluster" and gave each its own subsection
 - Added print(result.head()) to the verification snippet
 - Split the troubleshooting block so search_filter advice no longer lives under "idle workers", and corrected the partitioning guidance to match how HATS catalogs actually work

All changes were validated end-to-end against a real Kubernetes 1.33 cluster (Dask Operator 2026.3.0, lsdb 0.9.0): both connection paths work, result.head() prints clean rows, the static NFS PV+PVC binds, and the VersionMismatchWarning behaviour documented in the new tip reproduces exactly. While testing I also surfaced two real-world gotchas worth documenting that are now in the doc: (1) PVCs stuck in Pending when there is no default StorageClass, and (2) manifests rejected by ResourceQuota rules that require limits.cpu.

olivialynn (approved):
This looks great! Apologies for the delay - feel free to merge whenever!

mithr4ndir:
Thanks Olivia! I don't have merge rights on the base repo, so flagging that this is ready whenever a maintainer has a moment to hit the button. The RTD red X is the pre-existing import_catalogs.ipynb assertion failure (happy to dig into this too if you guys want me to :) ) on main, not this PR.

nevencaplar:
@mithr4ndir I sent you an invite to hats-friends group, that should allow you to push a bit green button.

PR #1330, opened 2026-04-07
Change `align_and_apply` to use map_partitions

### LINKED ISSUE #300: Explore dask expressions support using `ddf.partitions` and `map_partitions` instead of delayed
Instead of using dask delayed to align and map over the partitions of our catalogs, we could try to use the `ddf.partitions` accessor to align the partitions as necessary and map_partitions over them. There are still questions over how we deal with empty partitions and divisions, but may be an approach to look into.

smcguire-cmu:
Use `partitions` selector and map_partitions to do align_and_apply instead of going to and from delayed.

Closes #300 

This has a few behavior changes from the implementation before, mostly due to the limitations of expressions vs delayed:

- non-existing partitions in a catalog now are passed as empty data frames to the ufunc instead of None. I do think this is a better design decision overall, and further refactoring work could be done to remove passing meta as we do in a few places. This did expose some bugs though, particularly in left join nested, which resulted in changing how pixel_search was written to fix this.
- Catalogs that are None still pass None types to the ufunc, but in doing this we no longer support completely empty catalogs (0 columns). There may have been other places this didn't work before though, and I don't think its a use case we care about.
- There are a lot more dask tasks in the graph to pass all the arguments in the right places.

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -875,7 +875,7 @@ class HealpixDataset:
         return self.search(OrderSearch(min_order, max_order))
 
     def pixel_search(
-        self, pixels: tuple[int, int] | HealpixPixel | list[tuple[int, int] | HealpixPixel]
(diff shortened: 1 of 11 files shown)


dougbrn (approved):
Nice, this looks great!

PR #1332, opened 2026-04-09
Delayed Streaming Implementation

dougbrn:
Made in collaboration with Claude

Improves streaming performance by performing up front graph optimization and task culling, switching to delayed objects. Generally I see that partitions_per_chunk ~2x-4x the number of client workers is best, so I liked it as a default, but it's not clear to me that we have access to a client that was defined elsewhere without directly including it in the CatalogStream setup call, so opted to leave it as a recommendation in the docs.


smcguire-cmu (approved):
Looks good!

PR #1339, opened 2026-04-23
Add pyarrow string doc section to dask tips

### LINKED ISSUE #555: Dask creates Arrow String Conversion tasks
Even though we use pyarrow types, dask still makes these PyarrowStringConversion tasks. Can we make it not do that?

smcguire-cmu:
Closes #555


hombit (commented):
Would that task be useful if people added string object columns with map_rows/map_partitions?

smcguire-cmu:
It doesn't actually get added after map_partitions, only after a from_* for creating a ddf. We could add that people should use pyarrow string dtype in a map_partitions/rows call though?

hombit:
@smcguire-cmu yes, can you please add something on that? I'm not sure what is the right recommendation, but we should at least make people aware of it.

smcguire-cmu:
I've added something on it, let me know what you think

hombit (approved):
Thanks!

PR #1346, opened 2026-04-28
Implementing a `to_lance` output for HealpixDatasets

drewoldag:
This PR adds a new module, `to_lance.py` that will iterate through a healpix dataset and write out the contents as a lance dataset. 

I have a couple practical questions: 
1. I have added `lancedb` as an optional dependency, `[lance]` in pyproject.toml. As a side effect, I had to update a lot of github workflow files to install that opt. dep. so the unit tests would pass. Should I leave this as is, or put lancedb as a core dep. and revert the changes to the CI workflows? 
2. In `to_lance.py:to_lance` I use some private functions (`catalog._ddf.to_delayed()`, `catalog._ddf_pixel_map.items()`, is there a public API that I should be using? 
3. I'm not super familiar with standard LSDB workflows, so I'm not sure if I've put this `to_lance` function on the correct parent class (`HealpixDataset`). Is this the correct place to put it? Are there additional classes that would be written out? I assume that a user will do something like:
 ```python
cat = lsdb.open_catalog(...)
# filter, select, xmatch, end up with a HealpixDataset typically?
cat.to_lance(...)
``` 
4. Also I have somehow broken documentation? 🤷

### CODE REVIEW

File: src/lsdb/io/to_lance.py [hunk 1]
```diff
@@ -0,0 +1,94 @@
...
+        df = delayed_partitions[partition_index].compute()
+        if len(df) == 0:
+            continue
+
+        pa_table = pa.Table.from_pandas(df, preserve_index=True)
+
+        if not wrote_any:
+            mode = "overwrite" if overwrite else "create"
+            lance.write_dataset(pa_table, path, mode=mode)
+            wrote_any = True
```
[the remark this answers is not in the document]
drewoldag: done

drewoldag:
I believe that I've broken the benchmarks due to the way that I've isolated the `lancedb` dependency. If I could get some feedback on how to proceed there, I'll be happy to fix up the benchmarks. I can add the optional dep. if the preference is to keep it separate, or I can move lancedb to the main list of dependencies and clean up as needed.

src/lsdb/io/to_lance.py
```diff
@@ -0,0 +1,116 @@
...
+            table = db.create_table(table_name, pa_table, mode=mode)
+        else:
+            table.add(pa_table)
+
+    if table is None:
+        raise RuntimeError("The output catalog is empty. No data was written to Lance.")
+
+    # TODO: Replace with appropriate logging message and level
+    print("Optimizing Lance dataset...")
+    table.optimize(cleanup_older_than=timedelta(0), delete_unverified=True)
```
delucchi-cmu: Does the optimization step have to happen? Could it be disabled with a kwarg?

src/lsdb/io/to_lance.py
```diff
@@ -0,0 +1,116 @@
...
+    for _, partition_index in tqdm(
+        pixel_partition_pairs,
+        desc="Writing to Lance",
+        disable=not progress_bar,
+    ):
+        df = delayed_partitions[partition_index].compute()
+        if len(df) == 0:
+            continue
+
+        pa_table = pa.Table.from_pandas(df.reset_index(), preserve_index=False)
```
delucchi-cmu, line 108: How does this work with nested columns? You can add a test that uses an existing nested table with the `small_sky_order1_nested_sources` catalog.

src/lsdb/io/to_lance.py
```diff
@@ -0,0 +1,116 @@
...
+    catalog: HealpixDataset,
+    *,
+    base_catalog_path: str | Path | UPath,
+    table_name: str = _TABLE_NAME,
+    overwrite: bool = False,
+    progress_bar: bool = True,
+) -> None:
+    """Writes a catalog to a Lance dataset.
+
+    All partitions are written as a single flat Lance dataset. Every column in
```
delucchi-cmu:
```suggestion
    All primary catalog partitions are written as a single flat Lance dataset. Every column in
```

src/lsdb/io/to_lance.py
```diff
@@ -0,0 +1,116 @@
...
+    the catalog — including the HEALPix spatial index — is preserved. The
```
delucchi-cmu:
```suggestion
    the catalog — including the HEALPix spatial index — is preserved. The margin catalog (if present) is not written to Lance. The
```

src/lsdb/io/to_lance.py
```diff
@@ -0,0 +1,116 @@
...
+    ImportError
+        If the ``lancedb`` package is not installed.
+    ValueError
+        If a dataset already exists at ``base_catalog_path`` and ``overwrite=False``.
+    RuntimeError
+        If the catalog is empty and no data is written.
+
+    Examples
+    --------
+    Export a catalog and open it with lancedb:
```
delucchi-cmu, line 61: This docstring will not automatically appear on docs.lsdb.io. You will need to add the method explicitly to `docs/reference/io.rst`

delucchi-cmu (commented):
1. let's treat this dependency how we treat other optional dependencies - matplotlib is a good example. you would:
  * add it to the pyproject.toml's `[full]` dependency set
  * this means you can remove from the workflows, as we install full appropraitely (and ALSO have tests to confirm that things still work when the optional dependencies are NOT installed)
  * if the docs refer to the dependency (which is the case for the examples in the docstrings), you can add to `docs/requirements.txt`
  * localize the import just to the methods that require it. 
  
  ```
  ## file-level imports, if you're providing type hints
  if TYPE_CHECKING:
    from matplotlib.figure import Figure
  ## method-level imports
  def method():
        try:
            # pylint: disable=import-outside-toplevel
            from matplotlib import pyplot as plt
        except ImportError as exc:
            raise ImportError(
                "matplotlib is required to use this method. Install with pip or conda."
            ) from exc
  ```
  since this is our preferred approach, we could probably disable the `import-outside-toplevel` linting. but that's for another day. and maybe some `## Adding depencencies` in the contributor guide.

2. Sean and Doug are hard at work deprecating `catalog._ddf` as we speak. So don't worry about it for now.
3. That's the right place to put it.
4. See 1.
5. Please add (or open issue to add) test to hats-cloudtests, so we can see if this will work to write directly to s3. I'm going to guess it doesn't, but that might be a shortcoming of lancedb?

camposandro:
Melissa's suggestion of moving lancedb to the full dependency set should fix the benchmarks too!

src/lsdb/io/to_lance.py
drewoldag: We certainly could disable this optimization with kwargs. I fully acknowledge that it takes a long time to optimize a lance dataset, but I _think_ it should default to doing the optimization, but I'm happy to inject some kwargs that will disable it.

src/lsdb/io/to_lance.py
drewoldag: I'll check on this and add a test.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -1461,6 +1461,58 @@ def write_catalog(
...
+        RuntimeError
+            If the catalog is empty.
+
+        Examples
+        --------
+        >>> import lsdb
+        >>> catalog = lsdb.read_hats("path/to/small_sky")  # doctest: +SKIP
+        >>> catalog.to_lance("/tmp/my_catalog.lance")  # doctest: +SKIP
+        """
+        from lsdb.io.to_lance import to_lance as _to_lance
```
drewoldag: I ended up having to do this after I moved the `import lancedb` into the function. It seems that it changed python's import ordering and was `io.to_lance` to shadow the `to_lance` function and was triggering a circular dependency since we import HealpixDataset to use as a type hint in to_lance.py.

Let me know if what I've done here doesn't align with best practices.

src/lsdb/io/to_lance.py
drewoldag: Added a couple tests for nested

src/lsdb/io/to_lance.py
drewoldag: Added a kwarg that folks can skip the optimization step.

drewoldag:
@delucchi-cmu For point 5 in your last message - I have created an issue here: https://github.com/astronomy-commons/hats-cloudtests/issues/145

Once we play around with `to_lance` a bit more and have some confidence that it's going to give us what we want, we can do the experimentation to make sure that tests writing to s3 work as expected.

src/lsdb/catalog/dataset/healpix_dataset.py
delucchi-cmu: I fiddled around with this locally, and didn't have trouble with changing this to use the `io.to_lance` invocation. Could be some cached state in your dev setup? 

I was also looking at how we handle it in `to_hats`, as there are no circular dependencies there, and couldn't find any other differences.

tests/lsdb/io/test_to_lance.py
```diff
@@ -0,0 +1,135 @@
...
+
+def test_to_lance_empty_catalog_raises(small_sky_catalog, tmp_path, monkeypatch):
+    """An all-empty catalog raises RuntimeError with an informative message."""
+    ds_path = tmp_path / "small_sky"
+
+    # Patch the pixel map to be empty so no partitions are iterated
+    monkeypatch.setattr(small_sky_catalog, "_ddf_pixel_map", {})
+
+    with pytest.raises(RuntimeError, match="The output catalog is empty"):
+        small_sky_catalog.to_lance(ds_path)
```
delucchi-cmu:
```suggestion
    cone_search_catalog = small_sky_catalog.cone_search(0, -80, 1)

    with pytest.raises(RuntimeError, match="The output catalog is empty"):
        cone_search_catalog.to_lance(ds_path)
```

delucchi-cmu: After the cone search, the catalog really is empty. This will be a little more future-proof.

tests/lsdb/io/test_to_lance.py
```diff
@@ -0,0 +1,135 @@
+import importlib
+import sys
+
+import pandas as pd
+import pytest
+
+lancedb = pytest.importorskip("lancedb")
```
delucchi-cmu: I have a slight preference for putting the skip statements within the `test_` methods, but I can't really justify it, so feel free to ignore this comment.

tests/lsdb/io/test_to_lance.py
drewoldag: Cool, updated.

tests/lsdb/io/test_to_lance.py
drewoldag: Moved this into the various `test_*`s that are using lancedb. Let me know if this aligns with your preference.

src/lsdb/catalog/dataset/healpix_dataset.py
drewoldag: I'll recreate my local environment and see if it's the case that something is in an odd state.

drewoldag: Hmm, I recreated my environment, and attempted removing this `from lsdb.io.to_lance...` line and just calling `io.to_lance` directly.

When I do that, almost all the tests fail with this:
```
>       io.to_lance(
            self,
            base_catalog_path=base_catalog_path,
            table_name=table_name,
            overwrite=overwrite,
            progress_bar=progress_bar,
            optimize_dataset=optimize_dataset,
        )
E       TypeError: 'module' object is not callable
```

If I then change the line to `io.to_lance.to_lance(...)` _and_ remove the `from lsdb.io.to_lance ...` line, then all the tests pass and the imports work. 

But I can't figure out why this is different than the behavior of the `to_hats(...)` call. Still investigating.

drewoldag: Ok, got it figured out, the first unit test in the suite was removing to_lance as a module from the imports to test that the expected assertion raises. This was causing cascading failures in the rest of the tests. It is resolved now.

delucchi-cmu (approved):

PR #1359, opened 2026-05-04
LSDB Operations Implementation

### LINKED ISSUE #1342: Implement LSDB Custom Graphs
This issue is an overview issue for the work needed to implement custom graph construction within LSDB. 

Design Doc: https://docs.google.com/document/d/1sJNtDn1g_NhRUFb0SMzgvYaYYbQySlkGRMw7Nwjth1s/edit?tab=t.0 

TO DO:

- [ ] Port separate implementation from https://github.com/lincc-frameworks/lsdb-custom-graphs into a branch on LSDB
- [ ] Create any needed operations for existing LSDB functions
- [ ] Refactor LSDB functions to use operations
- [ ] Resolve unit test issues and any CI-based docs issues
- [ ] Validate static docs content
- [ ] Any other edge cases to test beyond our CI/docs?
- [ ] Remove lsdb.nested code as it will be made unnecessary by this

dougbrn:
Closes #1342 

### Motivation
This PR implements a sizable rewrite of LSDB's backend relationship with Dask. Where previously, LSDB was built directly on top of Dask Dataframe and much of LSDBs functionality relied on dask primitives to dictate graph construction (for the most part, in some cases we moved to custom delayed-style graphs for better control). With LSDB Operations, LSDB takes complete control of graph construction through it's own custom set of `Operations`.

The driving motivation being that we have observed graph sizes/construction time to be a clear limiter on scalability, and when relying on Dask for graph construction we have several bottlenecks/fall over modes related to tasks getting too much information stored in them and pre-culled graph sizes being large. However, in principle we know exactly what the optimized graph should look like for nearly all LSDB workflows. Workflows either involved spatial partition matching, which we know exactly which partitions should be interwoven, or they are simple map_partitions style work. LSDB Operations are designed to arrive at this optimal graph directly.

The exact results depend on the workflow, but as a rough estimate we expect a ~3-5x improvement in number of graph tasks and graph size in memory. This can have downstream speedups in actual computation, simply by avoiding the Dask per task overhead (~1ms) through reducing the total number of tasks.

### API Changes
The major change is that the `._ddf` property, which accessed the underlying Dask Dataframe directly is no longer available, because there is no longer an underlying Dask DataFrame! 

The public API is for the most part untouched, but there are exceptions:

Additions:
* Adds an explicit `Catalog.to_dask_dataframe` method as a replacement for the loss of ._ddf
* Adds `Catalog.exploded_columns` (previously only available through ._ddf)

Removals:
* `Catalog.merge` is removed as it was just a wrapper for Dask's merge, use it directly in Dask `Catalog.to_dask_dataframe().merge`
* remove HealpixDataset.get_partition_index

Behavior Changes:
* map_partitions meta has been expanded to accept more input data types (series or dict), and coerces results more heavily into dataframes (which return as catalogs)
* HealpixDataset.get_partition returns HealpixDataset not ddf
* .partitions supports healpix pixels as well as indices
* to_delayed defaults to optimize=false instead of true
* HealpixDataset.partitions is now iterable
* HealpixDataset.sample no longer computes multiple times
* HealpixDataset.random_sample doesn’t use delayed
* prune_empty_partitions no longer has persist option
* Column selection: Selecting a single column (e.g. catalog["my_column"]) now returns a catalog rather than a dd.series, however single column syntax is still usable for mask-style filtering (e.g. catalog[catalog["my_column"] == True]
* No longer have a nest accessor (for nested columns) available from a catalog column

### Technical State
Unit test suite has been fully adapted to operations and is now all passing. Mostly only mypy failures now, some of which actually currently exist in the codebase as this PR introduces a pandas typing dev dependency that gives mypy additional typing context that it previously lacked.

Docs have been reviewed (one pass) both for code changes (._ddf replacement) and language consistency with operations. Static notebooks were checked for obvious language/code changes, **but have not been rerun.**

The operations branch was tested on a focus week sprint, which is useful to mention just to illustrate it's been tested outside of the scope of our CI.

### Future Considerations
* The recently added LSDB Streaming implementation had some nice performance gains over the current LSDB (particularly by avoiding the large pre-culled graph sizes present in pre-operations LSDB), but potentially less so when compared against operations. While this is nice, we should look further into optimizing streaming for operations.
* Query optimization: We have enough control over the graphs to be able to retroactively optimize our graph execution, for example loading all columns but only providing a workflow that works with 1 of them. We didn't want to include this in our initial implementation but it is worth investigating in the near future.
* Series Integration: We currently sort of replicate some series behavior such as column selection, mask filtering, or map_partitions returning a scalar, with single column data frames. We may want to consider switching to either our own Operations backed Series or integrating with Dask Series. Some exploration on this has been done on the [sean/series](https://github.com/astronomy-commons/lsdb/tree/sean/series) branch


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -191,19 +248,300 @@ def _repr_html_(self):
...
+
+    def _create_updated_dataset(
+        self,
+        op: Operation | None = None,
+        hc_structure: HCHealpixDataset | None = None,
+        updated_catalog_info_params: dict | None = None,
+    ) -> Self:
+        """Creates a new copy of the catalog, updating any provided arguments
+
+        Shallow copies the ddf and ddf_pixel_map if not provided. Creates a new hc_structure if not provided.
```
delucchi-cmu: Nope.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -191,19 +248,300 @@ def _repr_html_(self):
...
+        updated_catalog_info_params : dict or None, default None
+            The dictionary of updates to the parameters of the hats dataset object's catalog_info
+
+        Returns
+        -------
+        Self
+            A new dataset object with the arguments updated to those provided to the function, and the
+            hc_structure metadata updated to match the new ddf
+        """
+        op = op if op is not None else self._operation
```
delucchi-cmu, line 296: Based on the old docstring, should this be a shallow copy, too?

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -191,19 +248,300 @@ def _repr_html_(self):
...
+        return self._create_updated_dataset(op=new_op)
+
+    def _check_single_column_for_boolean_op(self, op_name: str) -> None:
+        if len(self.columns) != 1:
+            raise ValueError(
+                f"Operator '{op_name}' requires a single-column catalog. "
+                "Use catalog['column_name'] to select a column first."
+            )
+
+    def _apply_comparison(self, op, other) -> Self:
```
delucchi-cmu, line 456: These comparison things could use some comments on motivation, usage, and gotchas.

src/lsdb/operations/lsdb_ops.py
```diff
@@ -0,0 +1,475 @@
...
+        """Get the catalog infos of the input catalogs"""
+        return [cat.hc_structure.catalog_info if cat is not None else None for cat in self.input_cats]
+
+    @property
+    def name(self) -> str:
+        """Return the name of the resulting AlignAndApply operation"""
+        op_names = ", ".join(op.name if op is not None else "None" for op in self.input_ops)
+        return f"AlignAndApply({funcname(self.func)}, {op_names})"
+
+    @functools.cached_property
```
delucchi-cmu, line 431: You knew about this? The whole time?

src/lsdb/operations/lsdb_ops.py
```diff
@@ -0,0 +1,475 @@
...
+        """Build the HealpixGraph from the Operation."""
+        input_ops = self.input_ops
+        graphs = [op.build() if op is not None else None for op in input_ops]
+        func = _verified(self.func, self.meta) if self.verify_meta else self.func
+        graph: dict = {}
+        pixel_key_map = {}
+        for g in graphs:
+            if g is not None:
+                graph = graph | g.graph
+        for i, all_pixels in enumerate(zip(self.output_pixels, *self.pixel_lists)):
```
delucchi-cmu, line 466: This looks very loop-loop-loopy, in a way that would make python really unhappy for constructing graphs for catalogs with 80k partitions. Is it?

delucchi-cmu (commented):
I'm only like 15% through looking at it, but this is enough to make me very happy: 

Image

There are just so many cases where the change cleans up existing interfaces. I'm sure there's sadness coming, so this is probably a nice place to leave it for today.

One major thing this PR needs is a longer description. We'll need to be clear about breaking changes, the motivation, what is NOT being implemented yet, etc. It's going to show up quite a bit in blame =D

src/lsdb/streams/catalog_streams.py
```diff
@@ -106,12 +106,14 @@ def __init__(
         # same full graph. We cull each to its own subgraph so that
         # client.compute() only sends the minimal per-partition graph to the
         # scheduler, avoiding O(N^2) graph transmission overhead.
-        raw_delayed = catalog.to_delayed(optimize_graph=True)
-        shared_graph = dict(raw_delayed[0].__dask_graph__())
+
+        # It's possible that the full build here practically limits the memory gains of streaming
+        # TODO: Follow up on this after custom graph construction
```
delucchi-cmu, line 111: Is this still a TODO? There are places elsewhere where we are computing just a single partition (but the fact that it's just over one partition is known earlier in the operation-construction-lifecycle).

src/lsdb/loaders/dataframe/margin_catalog_generator.py
```diff
@@ -169,18 +170,16 @@ def _generate_dask_df_and_map(
 
         Returns
         -------
-        tuple[nd.NestedFrame, dict[HealpixPixel, int], int]
-            Tuple containing the Dask Dataframe, the mapping of margin HEALPix
-            to the respective partitions and the total number of rows.
+        tuple[Operation, dict[HealpixPixel, int], int]
```
delucchi-cmu:
```suggestion
        tuple[Operation, int]
```

src/lsdb/io/schema.py
```diff
@@ -1,21 +1,21 @@
 from __future__ import annotations
 
-import dask.dataframe as dd
+import pandas as pd
 import pyarrow as pa
 
 
-def get_arrow_schema(ddf: dd.DataFrame) -> pa.Schema:
-    """Constructs the pyarrow schema from the meta of a Dask DataFrame.
+def get_arrow_schema(df: pd.DataFrame) -> pa.Schema:
```
delucchi-cmu, line 7: Don't worry about this with the operations PR necessarily, but while I'm here:

This feels like a really silly module. It wraps a single line of code. And I'm curious how many times this ends up getting called, and if we should just have the arrow schema be a property of the catalog instead.

src/lsdb/io/to_hats.py
```diff
@@ -65,7 +66,12 @@ def perform_write(
         at the specified order.
     """
     if len(df) == 0:
-        return 0, SparseHistogram([], [], histogram_order)
+        histogram = SparseHistogram([], [], histogram_order)
+        write_histogram(histogram, base_catalog_dir, hp_pixel)
+        write_done_pixel(base_catalog_dir, hp_pixel)
+        return pd.DataFrame({"count": [0], "histogram": [histogram]})
+    pixel_dir = hc.io.pixel_directory(base_catalog_dir, hp_pixel.order, hp_pixel.pixel)
+    hc.io.file_io.make_directory(pixel_dir, exist_ok=True)
```
delucchi-cmu: You don't need to make the new directory, if calling `new_pixel_catalog_file` - it takes care of that, so long as `create_dirs=True`, which it is by default.

src/lsdb/io/to_hats.py
```diff
@@ -521,29 +527,29 @@ def write_partitions(
         as well as the array with the sparse count histograms.
     """
     base_catalog_dir_fp = hc.io.file_io.get_upath(base_catalog_dir_fp)
-    results, pixels = [], []
-    partitions = catalog._ddf.to_delayed()
+    pixels = []
     existing_pixels_set = set(existing_pixels) if existing_pixels is not None else set()
 
-    for pixel, partition_index in catalog._ddf_pixel_map.items():
+    for pixel in catalog.get_healpix_pixels():
```
delucchi-cmu: pixels = [p for p in catalog.get_healpix_pixels() if p in existing_pixel_sets]

src/lsdb/io/to_hats.py
```diff
@@ -521,29 +527,29 @@ def write_partitions(
...
-                npix_parquet_name,
-                **kwargs,
-            )
-        )
         pixels.append(pixel)
 
-    if len(results) > 0:
-        results = dask.compute(*results)
-        counts, histograms = list(zip(*results))
+    if len(pixels) > 0:
```
delucchi-cmu: Kinda nitty, but maybe instead:

```
if len(pixels) == 0:
  return [],[],[]

write_cat = ... 
```

Reduces the indentation for interesting things, and makes it clearer that there's nothing more to do if there are no pixels to process.

src/lsdb/loaders/dataframe/from_dataframe_utils.py
```diff
@@ -1,17 +1,20 @@
...
 from hats.pixel_math.spatial_index import SPATIAL_INDEX_COLUMN
 
-import lsdb.nested as nd
-from lsdb.dask.divisions import get_pixels_divisions
+from lsdb.operations.lsdb_ops import FromHealpixMap
+from lsdb.operations.operation import Operation
 
 
-def _generate_dask_dataframe(
+def _pixel_df(pixel, df=None):  # pylint: disable=unused-argument
```
delucchi-cmu, line 11: Is this used anywhere?

src/lsdb/operations/functions/concat_catalog_data.py
```diff
@@ -6,18 +6,18 @@
...
-    construct_catalog_args,
     filter_by_spatial_index_to_margin,
     filter_by_spatial_index_to_pixel,
     get_aligned_pixels_from_alignment,
     get_healpix_pixels_from_alignment,
 )
-from lsdb.types import DaskDFPixelMap
+from lsdb.operations.operation import Operation
+
+# from lsdb.types import DaskDFPixelMap
```
delucchi-cmu:
```suggestion

```

src/lsdb/operations/lsdb_ops.py
```diff
@@ -0,0 +1,475 @@
...
+        return npd.NestedFrame({k: pd.Series(dtype=v) for k, v in meta})
+    raise ValueError(
+        f"meta must be a DataFrame, dict of {{name: dtype}}, or list of (name, dtype) tuples, got {type(meta)}"  # pylint: disable=line-too-long
+    )
+
+
+class MapPartitions(Operation):
+    """An Operation that applies a function to each partition of a HealpixGraph."""
+
+    class_func = None
```
delucchi-cmu, line 218:
```suggestion
    class_func = None
    """Subclasses with a predictable function supply this, and a custom user-function cannot be applied in that case"""
```

src/lsdb/operations/lsdb_ops.py
```diff
@@ -0,0 +1,475 @@
...
+    class_include_pixels: bool | None = None
```
delucchi-cmu, line 220:
```suggestion
    class_include_pixels: bool | None = None
    """Subclasses that determine their own pixels will supply this, and a custom user-defined set of pixels cannot be supplied in that case."""
```

tests/lsdb/operations/test_ops.py
```diff
@@ -0,0 +1,313 @@
...
+def test_coerce_to_meta_scalar_infers_dtype_from_type(scalar, expected_dtype):
+    meta = _coerce_to_meta(scalar)
+
+    assert len(meta) == 0
+    assert list(meta.columns) == ["result"]
+    assert meta["result"].dtype == expected_dtype
+
+
+def test_coerce_to_meta_unsupported_scalar_type_raises_type_error():
+    # KNOWN ISSUE: an object of a type pandas doesn't recognize as a dtype
```
delucchi-cmu: So if the dataframe contains a row with `HealpixPixel`, we'll get a type error? I like to do that kinda thing A LOT in my homework pipelines, when there's some summary statistics I'm creating for each pixel. Or am I misreading this note?

src/lsdb/operations/lsdb_ops.py
smcguire-cmu: I actually only found it when I was looking through dask expressions where they use it there. I have been using it a lot since

src/lsdb/operations/lsdb_ops.py
smcguire-cmu: Short answer is I think it's fine, and definitely better than what dask does already. 

Trying to make dictionaries like this there's really no easy way to vectorize it since it's all python datatypes, and unlike something like pandas map_rows, we're not constructing any expensive new data types per loop. We have tested it with >80k partitions and we are faster than dask for it, so while there might be optimizations we can do with graph libraries that are written in C or something like that, I think it's already fast enough, and already a big improvement over dask.

src/lsdb/catalog/dataset/healpix_dataset.py
smcguire-cmu: The previous code didn't actually shallow copy the ddf, I'm not sure why the docstring has that in it. Perhaps it's meant as in the Catalog object is shallow copied and reuses the ddf? But yeah the docstring should be changed.

src/lsdb/io/schema.py
smcguire-cmu: I agree that I don't like the schema being maintained in both the meta of the lsdb catalog and the property of the hats catalog. I think we should probably do something along the lines of just setting the hats schema to None when an lsdb catalog modifies it, or even just getting rid of it now that we have `original_schema` property on the hats catalog.

src/lsdb/loaders/dataframe/from_dataframe_utils.py
dougbrn: It's used just below: https://github.com/astronomy-commons/lsdb/blob/operations/src/lsdb/loaders/dataframe/from_dataframe_utils.py#L36

And it seems like (speaking a bit for @smcguire-cmu ) it's used as a compatibility piece for hats' FromHealpixMap

src/lsdb/streams/catalog_streams.py
dougbrn: I think this is still TODO, yeah. I added a related note to the future considerations section in the main PR text, but generally the graph build up front might make streaming not offer much improvement compared to operations in the way that it was efficient compared to pre-operations LSDB. I think this deserves a bigger investigation (I should make a ticket for it once operations goes in), but I would argue it's decoupled from the initial operations scope as it does work as implemented currently, and we've seen full graphs of DP2 be workable now as a baseline.

src/lsdb/io/to_hats.py
dougbrn: good streamlining, just needs to be `if p not in existing_pixel_sets`

tests/lsdb/operations/test_ops.py
dougbrn: This test is for `_coerce_to_meta`, which is used at the end of a map_partitions call to coerce the result structure into a dataframe. So if you return something with a HealpixPixel intended as a column, then that should be okay. But if you have a map_partitions call that returns a HealpixPixel, then that would fail and you would need to wrap it in a supported return type (like {"pixel_result": my_HealpixPixel}. Does that seem reasonable?

src/lsdb/loaders/hats/read_hats.py
```diff
@@ -424,42 +426,35 @@ def _load_dask_meta_schema(hc_catalog, config) -> npd.NestedFrame:
...
+def _load_operation(catalog: HCHealpixDataset, config) -> Operation:
     """Load Dask DF from parquet files and make dict of HEALPix pixel to partition index"""
     pixels = catalog.get_healpix_pixels()
     ordered_pixels = np.array(pixels)[get_pixel_argsort(pixels)]
-    divisions = get_pixels_divisions(ordered_pixels)
-
-    # Assign query parameters before a potential change of `config.columns`
-    # by _load_dask_meta_schema call.
+    dask_meta_schema = _load_dask_meta_schema(catalog, config)
+    index_column = dask_meta_schema.index.name
```
nevencaplar: It seems that you are creating index_column in this line and then repeating the same call 6 lines lower, in line 440.

dougbrn: good catch, removed the duplicate lines

src/lsdb/io/schema.py
smcguire-cmu: Made an issue for it #1443

src/lsdb/loaders/dataframe/from_dataframe_utils.py
smcguire-cmu: Yeah that's right, it's just a helper method to fit the signature expected by `FromHealpixMap`

tests/lsdb/operations/test_ops.py
smcguire-cmu: Could we coerce any unsupported dtype to an object column? I think that's what dask map_partitions would do currently and make it a series with a single element per partition, and I agree it's something I use currently

src/lsdb/streams/catalog_streams.py
dougbrn: Turned this into an issue: https://github.com/astronomy-commons/lsdb/issues/1444

tests/lsdb/operations/test_ops.py
dougbrn: Gotcha, that seems like what we want. I've updated the method to now have fallbacks to the np.object type whenever it tries to cast the type as a pandas type and fails.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -77,22 +84,72 @@ def __init__(
...
+        self._operation = operation
         self.hc_structure = hc_structure
         self.loading_config = loading_config
-        self._ddf_pixel_map = ddf_pixel_map
 
     def __repr__(self):
-        return self._ddf.__repr__()
+        # TODO: This changes to just be the operation name, not the lazy df view.
+        # Should we add a df like repr?
+        return self._operation.__repr__()
```
smcguire-cmu, line 101: Do we have an issue to improve this? Currently it is just the str repr e.g `MapPartitions(FromHealpixMap(...`, it might be nice to also include the columns or some kind of DF representation like dask df.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -191,19 +248,340 @@ def _repr_html_(self):
...
+    def map_partitions(
+        self,
+        func: Callable[..., npd.NestedFrame],
+        *args,
+        meta: pd.DataFrame | pd.Series | dict | Iterable | tuple | None = None,
+        include_pixel: bool = False,
+        compute_single_partition: bool = False,
+        partition_index: int | HealpixPixel | None = None,
+        **kwargs,
+    ) -> Self | dd.Series:
```
smcguire-cmu: We don't actually ever return a dd.Series anymore, right? If it's always a catalog we can remove the overloads and change the type hints to remove the types within the `Callable` from func and remove the `dd.Series` return type.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -191,19 +248,340 @@ def _repr_html_(self):
...
+            Additional keyword args to pass to the function. These are passed to the Dask DataFrame
+            `dask.dataframe.map_partitions` function, so any of the dask function's keyword args such as
+            `transform_divisions` will be passed through and work as described in the dask documentation
+            https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.map_partitions.html
+
+        Returns
+        -------
+        Self or dd.Series
+            A new catalog with each partition replaced with the output of the function applied to the original
+            partition. If the function returns a non dataframe output, a dask Series will be returned.
```
smcguire-cmu: Same return type issue as above

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -191,19 +248,340 @@ def _repr_html_(self):
...
+
+        new_op = MapPartitions(self._operation, func, *args, meta=meta, include_pixel=include_pixel, **kwargs)
+        new_cat = self._create_updated_dataset(op=new_op)
+        if not isinstance(new_op.meta, pd.DataFrame):
+            warnings.warn(
+                "output of the function must be a DataFrame to generate an LSDB `Catalog`. "
+                "`map_partitions` will return a dask object instead of a Catalog.",
+                RuntimeWarning,
+            )
+            return new_cat.to_dask_dataframe()
```
smcguire-cmu: How does this work with our meta inferencing logic? Is this reachable?

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -191,19 +248,340 @@ def _repr_html_(self):
...
+        hc_structure = self._create_modified_hc_structure(
+            hc_structure=hc_structure, updated_schema=updated_schema, **updated_catalog_info_params
+        )
+        return self.__class__(op, hc_structure, loading_config=self.loading_config)
+
+    def _apply_partitionwise_operation(
+        self, op_class: type[MapPartitions], func, *args, meta=None, **kwargs
+    ) -> Self:
+        new_op = op_class(self._operation, func, *args, meta=meta, **kwargs)
+        return self._create_updated_dataset(op=new_op)
```
smcguire-cmu, line 318: This is used in __get_item__ but not updated in the Catalog subclass, so the margin will never be updated. We should override this with margin logic in the subclass, and potentially change map_partitions to use it so we don't have to overwrite map_partitions in Catalog.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -191,19 +248,340 @@ def _repr_html_(self):
...
+            hc_structure = self.hc_structure.__class__(
+                catalog_info=self.hc_structure.catalog_info,
+                pixels=[pixel],
+                snapshot=self.hc_structure.snapshot,
+            )
+            return self.__class__(output_op, hc_structure)
+
+        new_op = MapPartitions(self._operation, func, *args, meta=meta, include_pixel=include_pixel, **kwargs)
+        new_cat = self._create_updated_dataset(op=new_op)
+        if not isinstance(new_op.meta, pd.DataFrame):
```
smcguire-cmu:
```suggestion
        new_cat = self._apply_partitionwise_operation(MapPartitions, func, *args, meta=meta, include_pixel=include_pixel, **kwargs)
        if not isinstance(new_cat, pd.DataFrame):
```

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -901,21 +1166,21 @@ def rename(self, columns: Renamer) -> Self:
...
             if new_ra_col != ra_col:
                 updated_params["ra_column"] = new_ra_col
         if dec_col in self.columns:
-            dec_col_df = self._ddf._meta[[dec_col]]
+            dec_col_df = self.meta[[dec_col]]
             new_dec_col = dec_col_df.rename(columns=columns).columns[0]
             if new_dec_col != dec_col:
                 updated_params["dec_column"] = new_dec_col
-        return self._create_updated_dataset(ddf=ndf, updated_catalog_info_params=updated_params)
+        new_op = MapPartitions(self._operation, lambda df: df.rename(columns=columns))
```
smcguire-cmu: This should use the self.map_partitions so the margin also gets updated in the Catalog

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -252,19 +251,15 @@ def load_catalog(self) -> Catalog:
...
         moc = self._generate_moc() if self.should_generate_moc else None
-        schema = self.schema if self.schema is not None else get_arrow_schema(ddf)
+        schema = self.schema if self.schema is not None else get_arrow_schema(op.meta)
         hc_structure = hc.catalog.Catalog(
             self.catalog_info, pixel_list, moc=moc, schema=schema, generate_snapshot=True
         )
-
-        # Recover NestedDtype (https://github.com/astronomy-commons/lsdb/issues/730)
-        if len(self.dataframe.nested_columns) > 0:
-            ddf = ddf.astype({col: self.dataframe[col].dtype for col in self.dataframe.nested_columns})
```
smcguire-cmu, line 265: Is this no longer needed?

src/lsdb/catalog/dataset/healpix_dataset.py
dougbrn: No issue yet, but I can make one. Towards the beginning of operations I tried to change this to a df-like repr, but it messed with some of the internal logic at that time where things were relying on the operations repr for naming. Not sure if that's still the case, but will make an issue!

dougbrn: https://github.com/astronomy-commons/lsdb/issues/1456

delucchi-cmu (approved):

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
dougbrn: I believe it's not needed because we're not ddf backed anymore

src/lsdb/catalog/dataset/healpix_dataset.py
dougbrn: Turned this into an issue, since I'm not quite sure I understand what you exactly want to have changed

src/lsdb/catalog/dataset/healpix_dataset.py
dougbrn: I don't think this really is reachable now with the expansion of meta checking, so I removed it

src/lsdb/catalog/dataset/healpix_dataset.py
dougbrn: #1460

PR #1362, opened 2026-05-05
Fixing up catalog unit tests

dougbrn:


src/lsdb/operations/lsdb_ops.py
```diff
@@ -201,20 +227,29 @@ def build(self) -> HealpixGraph:
...
-            else:
-                result = func(df, *args, **kwargs)
-            return _coerce_to_frame(result)
+        def wrapped_func(df, _partition_index, *args, **kwargs):
+            try:
+                if include_pixel:
+                    # pixel is already injected as first arg by the include_pixel path
+                    result = func(df, *args, **kwargs)
+                else:
+                    result = func(df, *args, **kwargs)
```
smcguire-cmu, line 236: We could probably remove this branch actually now that we're doing the pixel insertion elsewhere

PR #1365, opened 2026-05-11
Use same partitioning strategy for lsdb and hats-import.

delucchi-cmu:
Introduced this regression with https://github.com/astronomy-commons/hats-import/pull/673

Causes assert failure in docs build.


delucchi-cmu:
Unexpected improvement in benchmarks, too!

olivialynn (approved):

PR #1368, opened 2026-05-11
Automatically Update Getting Started LSDB Release Version

### LINKED ISSUE #908: Docs: pin the latest lsdb version in the installation snippets
E.g., `pip install lsdb>=0.6.0` (where 0.6.0 is the latest release) instead of `pip install lsdb`. This helps fail fast in cases where the installation defaults to an older release for some reason.

dougbrn:
Closes #908 

Approach is to ping pypi for the latest version, in the event that that ping fails, falls back to a known static version.


docs/conf.py
```diff
@@ -81,3 +83,24 @@
...
+def _latest_lsdb_version():
+    # Allow offline/sandboxed builds to fall back gracefully
+    try:
+        with urllib.request.urlopen("https://pypi.org/pypi/lsdb/json", timeout=10) as r:
+            return json.load(r)["info"]["version"]
+    except Exception:
+        # In the event that pypi is down, default to a known version.
+        # This will get increasingly out of date, but the odds of a PyPI
+        # outage should be low.
+        return "0.9.0"
```
delucchi-cmu: Instead of hard-coding, could we access the git tags and find the most recent release tag?

dougbrn: Hmm, alright I think I have the tags working. I think we still prefer the pypi result obviously for the best representation of what a user could grab, but the tags can be the next solution. I still think the hard-coding is worth it "just in case", where admittedly I'm not quite sure what it would take to get there. But it's pretty trivial to have

dougbrn:
Getting unit test failures, but I have no idea why anything changed here would result in that

Edit: Think I just was first to test a breakage due to this: https://github.com/lincc-frameworks/nested-pandas/commit/7511210e15248c405f55b83cfc45a3fa876548d6

delucchi-cmu (approved):

PR #1383, opened 2026-05-14
Add local directory to rsp clients

### LINKED ISSUE #1311: Rubin DP docs: recommend `/deleted-sundays` for tmp and outputs
We should recommend RSP users to use `/delete-sundays/${USER}` for large outputs and for Dask temporary data. We can start with
https://docs.lsdb.io/en/latest/tutorials/pre_executed/rubin_dp1.html
https://docs.lsdb.io/en/latest/tutorials/pre_executed/dp1-gaia-epoch-prop.html

smcguire-cmu:
Closes #1311


hombit (commented):
Could we do `os.environ("USER")`?

hombit (approved):

PR #1384, opened 2026-05-14
Add collections section to margin docs

### LINKED ISSUE #1172: Update documentation pages about margins with information about collections
https://docs.lsdb.io/en/latest/tutorials/margins.html - our margin documentation needs to be updated to reflect how many catalogs are collections now, and margins are there by default (but not always!)

smcguire-cmu:
Closes #1172

### DIFF
--- a/docs/tutorials/margins.ipynb
+++ b/docs/tutorials/margins.ipynb
@@ -30,11 +30,13 @@
     "collapsed": false
    },
    "source": [
-    "## 1. Loading a Catalog and Margin\n",
+    "## 1. Loading Catalogs with Margins\n",
     "\n",
-    "The margin cache is stored as a separate HATS catalog with the same partitioning as the main catalog.\n",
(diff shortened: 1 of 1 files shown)


docs/tutorials/margins.ipynb
```diff
@@ -47,16 +49,38 @@
...
+    "ztf.margin"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Loading a Catalog explicitly specifying a Margin Cache\n",
+    "\n",
+    "If you're not using a Catalog Collection, you can also specify the margin cache when opening a catalog. Margin caches are stored as a separate HATS catalogs with the same partitioning as the main catalogs. To load a catalog explicitly with a margin cache, we set the `margin_cache` parameter with the path of the catalog's margin catalog."
```
camposandro:
```suggestion
    "If you're not using a Catalog Collection, you can also specify the margin cache when opening a catalog. Margin caches are stored as separate HATS catalogs with the same partitioning as the main catalogs. To load a catalog explicitly with a margin cache, we set the `margin_cache` parameter with the path of the catalog's margin catalog."
```

camposandro (approved):

PR #1397, opened 2026-05-19
Crossmatch Types Documentation page

### LINKED ISSUE #1352: Add how-to guide explaining inner vs left crossmatch with example
The `crossmatch()` method supports two join types via the `how` parameter ('inner' and 'left'), but there is no user-facing guide explaining the difference, when to choose each, or what the output looks like in practice. 

**Context**

- Why needed: The default `how='inner' `silently drops all left-catalog sources with no match in the right catalog. 
- Current workflow: User calls `catalog_a.crossmatch(catalog_b, radius_arcsec=...`) and later discovers their result has fewer rows than catalog_a, with no explanation of why.
- Pain point: No documentation explains that catalog_a is the "left" catalog, that `how='inner'` (the default) keeps only matched pairs, and that how='left' retains all rows from catalog_a with NaN-filled right-catalog columns for unmatched sources.

**Goals** 

- A new how-to guide exists (e.g., docs/tutorials/types_of_crossmatch.ipynb) with:
-  A plain-language explanation of what "left" and "inner" mean in the context of sky crossmatching 
-  Explanation of which catalog is "left" and which is "right" when calling `catalog_a.crossmatch(catalog_b, ...)`
-  A concrete worked example using small synthetic catalogs showing: row counts for each join type, NaN columns in left join output for unmatched sources, and when you'd choose each approach
-  A note that the default is 'inner' and a recommendation of when to prefer 'left'
-  The how parameter docstring in` Catalog.crossmatch()` is updated to briefly describe the behavioral difference (not just list valid values)
- The existing [crossmatching.ipynb](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) tutorial links to or references the new guide

At the moment, just add the new guide in the TOC below existing crossmatching tutorial, as we will reorganize crossmatching section in a separate effort.

smcguire-cmu:
Closes #1352


nevencaplar (approved):
Looks great. Literally the only thing I would change is to push import statement lower, i.e., after opening the section ``Import Catalogs``, to keep intro clean from the code.

Image

PR #1399, opened 2026-05-21
Update logos and "Data Access" page

### LINKED ISSUE #1394: Update lsdb doc page describing data.lsdb.io
Now that Sandro has updated data.lsdb.io, the doc page describing it (with a screenshot of the old layout) is a little out of sync.

The new doc page may want to include:
- a new screenshot (if any screenshot at all)
- rewritten explanation of the hosting region (I think we now just break these into different pages, as if they were different catalogs--because sometimes, they are slightly different between data provider...)
- mention the new "View Skymaps" button, maybe also include a screenshot/brief explanation of the options available on that page
- maybe a quick explanation of the new "External Data Mirrors" menu in the top right

camposandro:
Update the LINCC/LSDB logos. Update the annotated screenshot of the "Data Access" page.


delucchi-cmu:
Does this close all the points in https://github.com/astronomy-commons/lsdb/issues/1394 ?

delucchi-cmu (approved):

Issue #1278, opened 2026-03-06
Create a canonical reference document for AI coding assistant for LSDB


nevencaplar:
Create canonical reference for AI coding assistants working on LSDB. Draw inspiration from [HYRAX_GUIDE.md](https://github.com/lincc-frameworks/hyrax/blob/main/HYRAX_GUIDE.md) but adapt it to LSDB architecture and conventions. 

The document should explain LSDB design principles, core abstractions (Catalog, partitions, HATS layout), API conventions, and performance constraints (partition-aware operations, avoiding unnecessary metadata scans or full dataset loads).

This guide should be the single shared reference for AI assistants and contributors. Tool-specific files (Copilot, Claude, etc.) should only contain overrides and reference this document.

Proposed filename: LSDB_GUIDE.md.

PR #1411, opened 2026-05-29
LSDB coding agents guide

camposandro:
Add LSDB Guide for coding agents. It would be valuable to have a guide focused on HATS as well; if we add it to the hats repo we can then link it from here. Part of #1278.

Rendered: https://lsdb--1411.org.readthedocs.build/en/1411/developer/lsdb-guide.html


dougbrn (approved):
Looks great!

PR #1412, opened 2026-06-01
Fix self-crossmatch failing due to identical default suffixes

### LINKED ISSUE #1275: Catalog self crossmatches fail due to suffix defaults

MayoKashyap:
Fixed self-crossmatch failing due to identical default suffixes and by adding ("", "_right")  to identical columns. 

fixes #1275

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -44,6 +44,8 @@ from lsdb.types import DaskDFPixelMap
 
 def _default_suffixes(left_name: str, right_name: str) -> tuple[str, str]:
     """Return the default pair of suffixes for left/right catalog names."""
+    if left_name == right_name:
+        return ("_l", "_r")
     return (f"_{left_name}", f"_{right_name}")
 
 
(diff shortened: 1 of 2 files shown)


hombit (commented):
Not related to the code, but I think we should think more about these suffixes. First, left column suffix changes ra/dec columns, which, as we discussed recently, not a nice thing to do by default. Second, single letter `_l` is not readable (is it 1 or l?). Third, We make people guess what these `_l`/`_r` mean.

I propose to go with `("", "_right")`, as polars do.

CC @delucchi-cmu @dougbrn

dougbrn:
I'm personally in favor of `("", "_right")`, I prefer avoiding column renaming when possible, but then if we have to, specifying "_right" fully gives someone a better chance of understanding what happened rather than just giving them "_r"

camposandro (approved):
Looks good, thanks!

PR #1435, opened 2026-06-22
Improve crossmatch coordinate column errors

### LINKED ISSUE #1393: Improve error when ra/dec column is not found in crossmatch.
Image

At the moment, when ra/dec is not found/avaliable for crossmatch, the error ``only`` identifies what the problem is (which is already quite good good). I propose adding a suggestion on how to fix it, i.e., something like ``Consider adding explicit names for ra/dec columns`` with a name/example for suitable args.

terminalchai:
Fixes #1393.

When a DataFrame crossmatch cannot infer its RA or Dec column, the error now shows the input-specific override (`left_args` or `right_args`) and the top-level alternative for catalogs that share a column name. Other `from_dataframe` validation errors remain unchanged.

The existing crossmatch regression test now verifies the RA and Dec suggestions.

Validation:
- `python -m pytest tests/lsdb/core/test_crossmatch_function.py -q` (9 passed)
- Black and isort checks on both changed files
- Pylint 10.00/10 on both changed files
- Mypy on both changed files
- Direct `lsdb.crossmatch` reproduction confirmed the new `right_args={'ra_column': '<column_name>'}` hint

### DIFF
--- a/src/lsdb/core/crossmatch/crossmatch.py
+++ b/src/lsdb/core/crossmatch/crossmatch.py
@@ -20,7 +20,18 @@ def _validate_and_convert_to_catalog(
         data_args["catalog_name"] = suffix if suffix else default_suffix
 
     # Convert the DataFrame to a Catalog.
-    data = from_dataframe(data, **data_args)
+    try:
+        data = from_dataframe(data, **data_args)
+    except ValueError as error:
+        for coordinate in ("ra", "dec"):
+            if str(error) == f"No column found for {coordinate}":
+                column_arg = f"{coordinate}_column"
+                raise ValueError(
+                    f"{error}. Specify the column with "
+                    f"{default_suffix}_args={{'{column_arg}': '<column_name>'}}, "
+                    f"or use {column_arg}=... when both inputs share the same column name."
+                ) from error
+        raise
     return data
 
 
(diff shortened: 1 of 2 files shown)


src/lsdb/core/crossmatch/crossmatch.py
```diff
@@ -20,7 +20,18 @@ def _validate_and_convert_to_catalog(
         data_args["catalog_name"] = suffix if suffix else default_suffix
 
     # Convert the DataFrame to a Catalog.
-    data = from_dataframe(data, **data_args)
+    try:
+        data = from_dataframe(data, **data_args)
+    except ValueError as error:
+        for coordinate in ("ra", "dec"):
+            if str(error) == f"No column found for {coordinate}":
```
camposandro, line 27: It would be ideal to raise this error on any `from_dataframe` call, whenever ra/dec columns are not specified, instead of doing it only within the context of crossmatching. Can we please also add the catalog's name to the error message? That way, if crossmatching two dataframes we will know which of the catalogs needs the explicit ra/dec columns specified.

nevencaplar:
No response from initial contributor, closing this pull. @terminalchai please reopen if you wish to proceed.

terminalchai:
Sorry for the delay here.

I tried to reopen this PR from my side, but GitHub is returning `Could not open the pull request` for the reopen request. I pushed the requested change to the same branch and opened a follow-up PR here: https://github.com/astronomy-commons/lsdb/pull/1451

The missing RA/Dec column validation now happens in the general `from_dataframe` path, and the error includes the catalog name so crossmatch identifies whether the `left` or `right` input needs explicit coordinate columns.

PR #1438, opened 2026-06-25
Add tutorial repo links

hombit:
I believe it would be nice to have all the tutorial links on the web page, even if some of the materials are out of date.


hombit:
Branch name is somehow harmful, re-creating...

PR #1439, opened 2026-06-25
Add tutorial links

hombit:
I believe it would be nice to have all the tutorial links on the web page, even if some of the materials are out of date.
Rendered (bottom of the page)
https://lsdb--1438.org.readthedocs.build/en/1438/tutorials.html

### DIFF
--- a/docs/tutorials.rst
+++ b/docs/tutorials.rst
@@ -25,7 +25,7 @@ materials from these events live in standalone GitHub repositories that you can 
 work through on your own:
 
 * `AstroAI 2026 LSDB Tutorial <https://github.com/lincc-frameworks/astroai2026-tutorial-lsdb>`_
-* `LSST-DA Regional Meeting at JHU 2025 — "From Transients to Cosmology" <https://github.com/lincc-frameworks/LSST-DA-meeting-JHU>`_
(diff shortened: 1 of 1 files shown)


docs/tutorials.rst
```diff
@@ -17,3 +17,18 @@ functionality.
...
    tutorial_toc/toc_debugging
 
+Workshop & Event Materials
+----------------------------------------------------------------------------------------
+
+We run hands-on LSDB tutorials at hackathons, workshops, and conferences. The
+materials from these events live in standalone GitHub repositories that you can clone and
+work through on your own:
+
+* `AstroAI Workshop 2026 LSDB Tutorial <https://github.com/lincc-frameworks/astroai2026-tutorial-lsdb>`_
```
nevencaplar: Can we get a little bit more explanation here. I think we should emphasize that users should look at the last two  tutorial repos, with a small mention that one is more ML, one is more Rubin data.

nevencaplar (commented):
I was just thinking today that we should do something like this!!!

docs/tutorials.rst
hombit: Hopefully addressed by [8c9b6a6](https://github.com/astronomy-commons/lsdb/pull/1439/commits/8c9b6a6c5a7cb40dfae1c3433a3e4c67ddae5ec3)

nevencaplar (approved):
Great! Looks good to me.

PR #1450, opened 2026-07-01
Operations-optimal Streaming

### LINKED ISSUE #1444: Investigate Streaming Graph Construction
CatalogStreams were introduced to offer an alternative graph construction method to avoid full graph builds for large workflows, which was particularly well-motivated by the larger size of constructed graphs prior to the LSDB operations refactor (#1359). 

CatalogStreams function appropriately with LSDB operations, but when I was porting over the logic, I wasn't entirely sure if it was using operations most effectively. Mainly, I'm concerned about the build call we do here: https://github.com/astronomy-commons/lsdb/pull/1359/changes#diff-00c0ff043bf349c6fb37b71cc0dacb1716166cef0853ab0b7d04e58b4d39d957R112

It's very possible that this means the peak graph size at a given time in memory is equivalent to a traditional compute call, since it's just call operation.build(). It would be good to spend some time looking at streaming and operations, and seeing if there are ways to more cleanly setup streaming with our new operations backend.

This was not critical to do as part of operations original implementation, as operations more directly addresses the problem that streaming was motivated by, but streaming has other benefits that are worth preserving.

dougbrn:
This should go in after #1359 is merged, pointed at main. Resolves #1444

### DIFF
--- a/src/lsdb/operations/lsdb_ops.py
+++ b/src/lsdb/operations/lsdb_ops.py
@@ -90,10 +90,13 @@ class FromHealpixMap(Operation):
     def healpix_pixels(self) -> list[HealpixPixel]:
         return self.pixels
 
-    def build(self) -> HealpixGraph:
+    def build(self, pixels: list[HealpixPixel] | None = None) -> HealpixGraph:
+        target = set(pixels) if pixels is not None else None
         graph = {}
         pixel_keys = {}
         for i, pixel in enumerate(self.pixels):
+            if target is not None and pixel not in target:
+                continue
             key = (self.key_name, i)
             map_kwargs = {k: v[i] for k, v in self.map_kwargs.items()} if self.map_kwargs is not None else {}
             func = _verified(self.func, self.meta) if self.verify_meta else self.func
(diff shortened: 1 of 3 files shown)


src/lsdb/operations/lsdb_ops.py
```diff
@@ -360,9 +365,11 @@ def is_reloadable(self) -> bool:
     def healpix_pixels(self) -> list[HealpixPixel]:
         return list(self.pixels)
 
-    def build(self) -> HealpixGraph:
+    def build(self, pixels: list[HealpixPixel] | None = None) -> HealpixGraph:
         """Build the HealpixGraph from the Operation."""
-        previous = self.base.build()
+        # intersect the caller's requested subset with a provided pixel filter
+        effective = [p for p in self.pixels if pixels is None or p in set(pixels)]
```
smcguire-cmu: I think the set should be outside of the list comprehension here, I think this way it's being converted to a set every iteration through the list.

src/lsdb/operations/lsdb_ops.py
```diff
@@ -453,28 +460,45 @@ def healpix_pixels(self) -> list[HealpixPixel]:
...
         """Build the HealpixGraph from the Operation."""
         input_ops = self.input_ops
-        graphs = [op.build() if op is not None else None for op in input_ops]
+
+        # resolve against a custom input pixel list
+        if pixels is not None:
+            requested = set(pixels)
+            output_indices = [i for i, p in enumerate(self.output_pixels) if p in requested]
+            input_pixel_subsets = [
+                list({pixel_list[i] for i in output_indices if pixel_list[i] is not None})
```
smcguire-cmu: Can this be a list comprehension?

smcguire-cmu (approved):
Looks good! A few minor suggestions before merging, but this is great overall thanks!

src/lsdb/operations/lsdb_ops.py
smcguire-cmu: (This should have been in the review not a separate comment btw, I just misclicked the button)

src/lsdb/operations/lsdb_ops.py
dougbrn: good call!

src/lsdb/operations/lsdb_ops.py
dougbrn: implemented

Issue #171, opened 2024-02-21
Assigning a column to a catalog and its margin


smcguire-cmu:
In working to make sure that catalog methods also apply to the margin catalog, assign seems like a harder case. If a user is assigning a series that is partitioned to fit the catalog's partitions, then to apply the same column to the margin we would need to do some sort of join or map/reduce to rearrange the column to fit with the margin's partitions. I think this would be fairly inefficient and break a lot of the extremely parallel operations we have.

Alternatively, we could leave this to the user, and accept the catalog and margin columns separately to assign, but this then leaves the work on the user.

hombit:
Maybe we can do it with `.eval()` and highlight it in the docs

PR #1464, opened 2026-07-09
Support dask series in operations for getitem and map_partitions

### LINKED ISSUE #1413: Operations - Series behavior
When working with single columns in operations catalogs, how much do we want them to look like series?

After discussion, we decided that filtering/slicing syntax such as `df[df["col"] > 10]` is important, and should be working before operations is released, but that other series behavior such as nested accessors and `dtype` is not required.

We may later want to investigate how feasible it is to implement more series-like behavior, but for now keeping things as a catalog is good enough as a first step

smcguire-cmu:
Supports returning a series for a single column `getitem`, filtering with a series for `getitem`, and map_partitions with a series output 

Closes #1413

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -45,8 +45,7 @@ from lsdb.core.search.region_search import (
 )
 from lsdb.io.schema import get_arrow_schema
 from lsdb.loaders.hats.hats_loading_config import HatsLoadingConfig
-from lsdb.operations.expressions import FromOperation, FromDaskExpression
-from lsdb.operations.functions.divisions import get_pixels_divisions
+from lsdb.operations.expressions import FromDaskExpression, FromOperation
 from lsdb.operations.functions.merge_catalog_functions import align_and_apply, concat_metas, make_meta
 from lsdb.operations.functions.partition_indexer import PartitionIndexer
 from lsdb.operations.lsdb_ops import (
@@ -433,7 +432,7 @@ class HealpixDataset:
             self._check_unloaded_columns([item])
             single_column_catalog = self[[item]]
             return single_column_catalog.to_dask_dataframe()[item]
-        elif isinstance(item, Sequence):
+        if isinstance(item, Sequence):
             self._check_unloaded_columns([col for col in item if isinstance(col, str)])
         meta = self.meta
         if isinstance(meta, npd.NestedFrame) and meta._is_key_list(item):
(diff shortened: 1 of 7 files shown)


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -371,9 +393,9 @@ def map_partitions(
 
         Returns
         -------
-        Self
+        Self or dd.Series
             A new catalog with each partition replaced with the output of the function applied to the original
-            partition.
+            partition. If the function returns a non dataframe output, a dask Series will be returned.
```
dougbrn, line 398: Do we have any support for assignment of series into catalog columns? Don't remember what the state of that was in LSDB prior to operations either, but generally a little worried about these return series not being that useful for users. Not blocking, but depending on the state could be a follow up issue

src/lsdb/operations/expressions.py
```diff
@@ -0,0 +1,89 @@
...
+        for i, pixel in enumerate(self.sorted_pixels):
+            graph_dict[(self._name, i)] = Alias((self._name, i), self.operation_graph.pixel_to_key_map[pixel])
+        return graph_dict
+
+    def _task(self, key: Key, index: int) -> Task:
+        raise NotImplementedError("FromOperation does not implement _task; use the _layer instead.")
+
+
+class FromDaskExpression(Operation):
+    """LSDB Operation to create a Dask Expression from an LSDB operation."""
```
dougbrn: nit: should the string here be "Create an LSDB operation from a Dask Expression"?

dougbrn (approved):
Looks good! Just a question and a nit, neither of which I consider blocking, though the nit would be nice to cleanup

dougbrn (changes requested):
blocking: doc fixes are needed

src/lsdb/catalog/dataset/healpix_dataset.py
smcguire-cmu: I think that we never actually allowed assigning a series outside of map_partitions. We've had #171 open for ages to support it that has been blocked on figuring out how to handle margins with series, but it might be worth revisiting allowing it without margins for now. But yeah I agree for this PR given that we didn't support it before I don't think it's worth adding it right now. And I think getting a series is still useful for users to compute and interact with as a series, keeping the same API as before, and it enables doing the filtering with series from a catalog.

src/lsdb/operations/expressions.py
smcguire-cmu: Yep, good catch thanks!

dougbrn (approved):

PR #1466, opened 2026-07-09
Update apply_partitionwise to work on margin in catalog

### LINKED ISSUE #1460: Fix operations margin updating in HealpixDataset
This is used in __get_item__ but not updated in the Catalog subclass, so the margin will never be updated. We should override this with margin logic in the subclass, and potentially change map_partitions to use it so we don't have to overwrite map_partitions in Catalog.

_Originally posted by @smcguire-cmu in https://github.com/astronomy-commons/lsdb/pull/1359#discussion_r3501741880_

smcguire-cmu:
Closes #1460


dougbrn (approved):
Nice, looks good!

PR #1473, opened 2026-07-12
Forward suffix_method/log_changes from lsdb.crossmatch() to Catalog.crossmatch() (#1470)

### LINKED ISSUE #1470: lsdb.crossmatch() doesn't forward suffix_method / log_changes to Catalog.crossmatch()
**Bug report**

The top-level `lsdb.crossmatch()` function has a hand-written, explicit signature that forwards only a *subset* of `Catalog.crossmatch()`'s parameters. Two params on the method are missing from both the function's signature and its delegating call, so a function-form caller cannot reach them:

- `suffix_method`
- `log_changes`

Because the function enumerates params manually (no `**kwargs` passthrough), any param added to the method must also be added here or it becomes silently unreachable.

**Concrete impact (`suffix_method`):** the method emits a `FutureWarning` telling users to set `suffix_method='all_columns'` / `'overlapping_columns'` — but a `lsdb.crossmatch(left, right)` caller has no way to comply. They must switch to the method form `left.crossmatch(right, suffix_method=...)`. The function always trips the warning with no escape hatch.

**Reproduce**
```python
import lsdb
region = lsdb.ConeSearch(ra=53.13, dec=-28.10, radius_arcsec=20)
a = lsdb.generate_catalog(100, 5, seed=53, search_region=region)
b = lsdb.generate_catalog(100, 5, seed=28, search_region=region)

lsdb.crossmatch(a, b, suffix_method="all_columns")   # TypeError: unexpected keyword argument
a.crossmatch(b, suffix_method="all_columns")         # works — no warning
```

**Root cause** — `lsdb/core/crossmatch/crossmatch.py` (`crossmatch()`, ~L28): the final `return Catalog.crossmatch(...)` passes `n_neighbors, radius_arcsec, min_radius_arcsec, algorithm, output_catalog_name, require_right_margin, how, suffixes` but not `suffix_method` or `log_changes`. Method def in `lsdb/catalog/catalog.py` (~L241).

**Fix** — add `suffix_method` and `log_changes` to the function signature and forward them in the delegating call.

**Related**
- #1389 (open) — removing the suffix FutureWarning by flipping defaults. Overlaps on the warning symptom but does *not* fix the forwarding gap; even after the default flips, a function-form caller still can't override the choice.
- #946 (closed) — a different facet (crossmatch *algorithm* kwargs / IDE discoverability), not the function→method forwarding gap.

**Environment:** lsdb 0.9.2

nevencaplar:
Fixes #1470 — `suffix_method` / `log_changes` were reachable from `Catalog.crossmatch()` but not from the top-level `lsdb.crossmatch()`.

The new params are appended after `left_args`/`right_args` rather than placed next to `suffixes`: the function accepts positional args (the method is keyword-only), so inserting mid-signature would rebind a positional `left_args`/`right_args`.

### DIFF
--- a/src/lsdb/core/crossmatch/crossmatch.py
+++ b/src/lsdb/core/crossmatch/crossmatch.py
@@ -28,6 +28,7 @@ def _validate_and_convert_to_catalog(
 def crossmatch(
     left: Catalog | npd.NestedFrame | pd.DataFrame,
     right: Catalog | npd.NestedFrame | pd.DataFrame,
+    *,
     ra_column: str | None = None,
     dec_column: str | None = None,
     n_neighbors: int | None = None,
@@ -40,6 +41,8 @@ def crossmatch(
     suffixes: tuple[str, str] | None = None,
     left_args: dict | None = None,
     right_args: dict | None = None,
+    suffix_method: str | None = None,
+    log_changes: bool = True,
 ) -> Catalog:
     """Perform a cross-match between two frames, two catalogs,
     a catalog and a frame, or a frame and a catalog.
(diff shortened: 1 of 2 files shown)


src/lsdb/core/crossmatch/crossmatch.py
```diff
@@ -40,6 +40,8 @@ def crossmatch(
     suffixes: tuple[str, str] | None = None,
     left_args: dict | None = None,
     right_args: dict | None = None,
+    suffix_method: str | None = None,
```
hombit, line 44: We should use more keyword-only args, and while it is technically a breaking chance, I still believe it is ok to change it in a patch release and introduce `*,` after `right` argument. If you disagree, let's at least add the new arguments after `*,`

hombit (approved):

src/lsdb/core/crossmatch/crossmatch.py
nevencaplar: Ok, added `*`, just after `right`. Is this what you had in mind?

PR #1474, opened 2026-07-13
Implement catalog view

### LINKED ISSUE #692: More information per column in catalog view instead of triple dots
At the moment the view of the lazy loaded catalog is empty and uninformative with triple dots (...) everywhere. 

With the aim to improve and have more informative catalogs representation, show for each column min and max instead of triple dots. Limit the number of digits shown in max and min value so that it is not longer than the name of the column on the top. Abandon this and go back to triple dots after the catalog has been touched.

camposandro:
Implements a lazy catalog view, similar to the one we had previously, before operations was merged.

I added the possibility of requesting partition/column statistics when opening a catalog: `open_catalog(..., show_statistics=True)`. The flag's default is set to False because this feature depends on either the `_metadata` or the `per_partition_statistics` parquet files that might (1) not be present in the catalog's directory, or (2) be too large for the users to download (especially from remote).

I tried to ensure that `SelectPixels` and `SelectColumns` are operations that preserve the pixel statistics. A catalog can therefore be loaded with a set of columns and a coarse search filter and its statistics will still be valid for the view. User-defined row `filters` and other operations (e.g. `map_partitions`) should invalidate statistics.

When statistics are not present, or are invalid, the Catalog representation defaults to a table view with ellipses ("...") as placeholders.

Closes #692.

### DIFF
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -280,6 +280,7 @@ class HealpixDataset:
         if len(cols) == 0:
             return pd.DataFrame([[]] * len(index), columns=cols, index=index)
         show_statistics = (
+            # In-memory catalogs have no loading config
             self.loading_config is not None
             and self.loading_config.show_statistics
             and self._operation.pixel_stats_preserved
@@ -295,12 +296,15 @@ class HealpixDataset:
             non_nested_cols = [c for c in cols if c not in self.nested_columns]
             if len(non_nested_cols) == 0:
                 return None
-            return self.per_partition_statistics(
-                use_default_columns=False,
-                exclude_hats_columns=True,
-                include_columns=non_nested_cols,
-                include_stats=["min_value", "max_value"],
-            )
+            # Only reached when pixel stats are preserved.
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", message=".*modified catalog.*")
+                return self.per_partition_statistics(
+                    use_default_columns=False,
+                    exclude_hats_columns=True,
+                    include_columns=non_nested_cols,
+                    include_stats=["min_value", "max_value"],
+                )
         except Exception:  # pylint: disable=broad-exception-caught
             logging.debug("Could not read partition statistics for catalog %s", self.name, exc_info=True)
             return None
--- a/src/lsdb/catalog/dataset/healpix_dataset.py
+++ b/src/lsdb/catalog/dataset/healpix_dataset.py
@@ -227,7 +227,20 @@ class HealpixDataset:
         return None
 
     def __repr__(self):
-        data = self._repr_data().to_string(max_rows=5, show_dimensions=False)
+        return self._format_text_repr(self._repr_data())
+
+    def _repr_html_(self):
+        return self._format_html_repr(self._repr_data())
+
+    def _repr_mimebundle_(self, **kwargs):  # pylint: disable=unused-argument
+        """Return both the text and HTML representations for notebook display."""
+        data = self._repr_data()
+        text = self._format_text_repr(data)
+        html = self._format_html_repr(data)
+        return {"text/plain": text, "text/html": html}
+
+    def _format_text_repr(self, data):
+        body = data.to_string(max_rows=5, show_dimensions=False)
         loaded_cols = len(self.columns)
         available_cols = len(self.all_columns)
         est_size = self.est_size()
@@ -238,13 +251,13 @@ class HealpixDataset:
         )
         return (
             f"lsdb Catalog {self.name}:\n"
-            f"{data}\n"
+            f"{body}\n"
             f"{loaded_cols} out of {available_cols} available columns in the catalog have been "
             f"loaded lazily, meaning no data has been read, only the catalog schema"
         ) + est_size_text
 
-    def _repr_html_(self):
-        data = self._repr_data().to_html(max_rows=5, show_dimensions=False, notebook=True)
+    def _format_html_repr(self, data):
+        body = data.to_html(max_rows=5, show_dimensions=False, notebook=True)
         loaded_cols = len(self.columns)
         available_cols = len(self.all_columns)
         est_size = self.est_size()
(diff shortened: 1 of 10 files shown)


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -243,11 +268,47 @@ def _repr_html_(self):
...
+        )
+        pixel_stats = self._get_repr_pixel_stats(cols) if show_statistics else None
+        if pixel_stats is None:
+            return pd.concat([_repr_data_series(s, index=index) for _, s in meta.items()], axis=1)
+        return _repr_data_min_max(meta, index, pixel_stats)
+
+    def _get_repr_pixel_stats(self, cols) -> pd.DataFrame | None:
+        """Fetch per-pixel min/max stats for display"""
+        try:
+            non_nested_cols = [c for c in cols if c not in self.nested_columns]
```
delucchi-cmu: `self.nested_columns` will only contain the nested columns that are actually loaded? e.g. if you've down-selected your columns, this isn't still going to show hidden columns?

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -243,11 +268,47 @@ def _repr_html_(self):
...
+            if len(non_nested_cols) == 0:
+                return None
+            # Only reached when pixel stats are preserved.
+            with warnings.catch_warnings():
```
delucchi-cmu: I would support adding a parameter to `hats.per_partition_statistics` (and `hats.aggregate_global_statistics`) like `warn_on_modified_catalog=True`, and set to `False` in this case.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -243,11 +268,47 @@ def _repr_html_(self):
...
+                warnings.filterwarnings("ignore", message=".*modified catalog.*")
+                return self.per_partition_statistics(
+                    use_default_columns=False,
+                    exclude_hats_columns=True,
+                    include_columns=non_nested_cols,
+                    include_stats=["min_value", "max_value"],
```
delucchi-cmu: The HTML representation is only going to show 5-10 partitions, right? Do you know which pixels that will be, when you make this call? If so, you could potentially speed things even further by limiting to those pixels with the `include_pixels` argument.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -231,8 +226,38 @@ def est_size(self) -> float | None:
...
 
+    def __repr__(self):
+        return self._format_text_repr(self._repr_data())
+
     def _repr_html_(self):
-        data = self._repr_data().to_html(max_rows=5, show_dimensions=False, notebook=True)
+        return self._format_html_repr(self._repr_data())
+
+    def _repr_mimebundle_(self, **kwargs):  # pylint: disable=unused-argument
+        """Return both the text and HTML representations for notebook display."""
```
delucchi-cmu: But why... can you give some kinda jupyter link on how this method would be called? I've just never seen it before.

src/lsdb/catalog/dataset/healpix_dataset.py
camposandro: Yes, that is correct - nested columns is a property of operations meta so it only shows which nested columns of the catalogs are loaded.

src/lsdb/catalog/dataset/healpix_dataset.py
camposandro: Thanks, I totally agree - opened [hats#705](https://github.com/astronomy-commons/hats/pull/705) to address that!

src/lsdb/catalog/dataset/healpix_dataset.py
camposandro: Yes! Actually, I will explain what brought me to this method.

I noticed that `__repr__` and `_repr_html_` were both getting called when displaying a Catalog in a Jupyter notebook. Since I was testing the rendering of the statistics and reading an _metadata file from remote that was taking a long while! It seems that we always had this issue but we never noticed since the catalog view with the placeholders was so quick to render. Interestingly, according to https://github.com/ipython/ipython/issues/9771, this is IPython's expected behavior.

I'm now using the `_repr_mimebundle_` method to render the two views (text and html) that Jupyter needs without having to compute the statistics twice, since that is expensive. I have got to admit, I didn't know about it either, but I saw it being used in this PR https://github.com/ibis-project/ibis/pull/10002 and decided to give it a try. The [method's documentation](https://ipython.readthedocs.io/en/stable/config/integrating.html#MyObject._repr_mimebundle_) states that when it is present, other `repr_` methods are ignored.

delucchi-cmu: Please add some more info in the docstring, so we don't take it out next year when we forget why it's there =D

dougbrn (commented):
Just looking through the PR, things looks reasonable to me. However, I'm still not quite sure what the actual impact of this is. Is there an A/B comparison somewhere for the html repr? I know the default repr was not good in operations, but unclear what has changed on the html side

src/lsdb/catalog/dataset/healpix_dataset.py
camposandro: Good point! We only show up to 4 pixels because the table has 5 rows and the header with the column types also counts as one.

camposandro:
> Just looking through the PR, things looks reasonable to me. However, I'm still not quite sure what the actual impact of this is. Is there an A/B comparison somewhere for the html repr? I know the default repr was not good in operations, but unclear what has changed on the html side

Thanks Doug! The html repr should match almost exactly the one we had before operations. This PR just adds some logic so that the table view (that previously just had placeholders) can show the per-pixel statistics of each column.

Screenshot 2026-07-14 at 10 01 40 PM

It's true that this information is already accessible via `Catalog.per_partition_statistics`, so this shouldn't be a must-have feature for DP2. I'm happy to hold off on it and open it to a wider discussion!

dougbrn (approved):
This looks good, and in the interest of making the v0.10.1 deadline I'm going to shepherd this through

PR #1451, opened 2026-07-01
Improve dataframe coordinate column errors

terminalchai:
Follow-up to #1435 after the review request in https://github.com/astronomy-commons/lsdb/pull/1435#discussion_r3460806814.

This moves the missing RA/Dec column error into the general `from_dataframe` path and includes the catalog name in the message, so the same validation applies outside crossmatch while crossmatch still reports whether the `left` or `right` input needs explicit coordinate columns.

Validation:
- `python -m pytest tests\lsdb\loaders\dataframe\test_from_dataframe.py::test_from_dataframe_finds_radec_columns tests\lsdb\core\test_crossmatch_function.py::test_ra_dec_columns_crossmatch -q`
- `python -m py_compile src\lsdb\loaders\dataframe\dataframe_catalog_loader.py src\lsdb\core\crossmatch\crossmatch.py`
- `python -m black --check src\lsdb\loaders\dataframe\dataframe_catalog_loader.py src\lsdb\core\crossmatch\crossmatch.py tests\lsdb\loaders\dataframe\test_from_dataframe.py tests\lsdb\core\test_crossmatch_function.py`
- `git diff --check`

PR #1476, opened 2026-07-13
Avoid pandas FutureWarning when concatenating an empty margin

### LINKED ISSUE #1471: Resolve Deprecation Warning in `concat_partition_and_margin`
When doing operations that invoke `concat_partition_and_margin`, like crossmatch, we get the following deprecation warning if the margin is empty (not None):
```
/opt/lsst/software/stack/conda/envs/lsst-scipipe-13.0.0-exact/lib/python3.13/site-packages/lsdb/dask/merge_catalog_functions.py:233: FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
  joined_df = pd.concat([partition, margin])
```

We should add an extra logic catch to check if the margin is empty and handle appropriately. A reproducer can be found here: https://github.com/lsst-so/linccf/blob/main/RSP/critical_fs.ipynb

vakudo:
Fixes #1471.

`concat_partition_and_margin` already returns early when the margin (or partition) is `None`, but a margin that is **present yet empty** still went through `pd.concat([partition, margin])`. Under pandas 2.x, concatenating an empty / all-NA frame raises:

```
FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes...
```

which surfaces during crossmatch and other operations that build an (empty) margin, as reported in the issue.

### Fix
Return the partition directly when the margin has no rows — an empty margin contributes nothing to the concatenation result — so both the `concat` and the warning are skipped. The `None` handling and the non-empty-margin path are unchanged. I kept the existing `partition is None` guard first so all previous None-handling behavior is preserved.

### Tests
Added two tests in `tests/lsdb/dask/test_merge_functions.py`:
- `test_concat_partition_and_margin_empty_margin_no_warning` — turns `FutureWarning` into an error and passes an empty margin with object-dtype columns (the shape that triggers the warning); it fails on the previous code and passes with the fix.
- `test_concat_partition_and_margin_nonempty_margin_concatenates` — guards that a non-empty margin is still concatenated.

`black`, `pylint` (10.00/10 with the repo rcfiles) and the `test_merge_functions.py` suite all pass on the changed files.

### DIFF
--- a/src/lsdb/operations/functions/merge_catalog_functions.py
+++ b/src/lsdb/operations/functions/merge_catalog_functions.py
@@ -225,10 +225,12 @@ def concat_partition_and_margin(
     npd.NestedFrame
         The concatenated dataframe with the partition on top and the margin on the bottom
     """
-    if margin is None:
-        return partition
     if partition is None:
         return margin
+    if margin is None or len(margin) == 0:
+        # An empty (but not None) margin contributes no rows. Skipping the concat
+        # avoids a pandas FutureWarning about concatenating empty/all-NA entries.
+        return partition
     joined_df = pd.concat([partition, margin])
     return npd.NestedFrame(joined_df)
 
--- a/src/lsdb/operations/functions/merge_catalog_functions.py
+++ b/src/lsdb/operations/functions/merge_catalog_functions.py
@@ -229,7 +229,7 @@ def concat_partition_and_margin(
         return margin
     if margin is None or len(margin) == 0:
         # An empty (but not None) margin contributes no rows. Skipping the concat
-        # avoids a pandas FutureWarning about concatenating empty/all-NA entries.
+        # avoids a pandas FutureWarning about concatenating empty frames.
         return partition
     joined_df = pd.concat([partition, margin])
     return npd.NestedFrame(joined_df)
--- a/tests/lsdb/dask/test_merge_functions.py
+++ b/tests/lsdb/dask/test_merge_functions.py
@@ -1,8 +1,38 @@
 import logging
+import warnings
+
+import nested_pandas as npd
+import pandas as pd
 
 from lsdb.operations.functions.merge_catalog_functions import (
+    concat_partition_and_margin,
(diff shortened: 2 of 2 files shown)


vakudo:
The pre-commit-ci failure here is the mypy/pandas-stubs drift that's currently failing on all open PRs (e.g. #1474, #1451) — the reported errors are in lines this PR doesn't touch.

hombit:
@vakudo thank you for the contribution. I think it is reasonable, however the comments say that it also covers all-NA case, but  I don't see how it actually does it.

src/lsdb/operations/functions/merge_catalog_functions.py
```diff
@@ -225,10 +225,12 @@ def concat_partition_and_margin(
     npd.NestedFrame
         The concatenated dataframe with the partition on top and the margin on the bottom
     """
-    if margin is None:
-        return partition
     if partition is None:
         return margin
+    if margin is None or len(margin) == 0:
+        # An empty (but not None) margin contributes no rows. Skipping the concat
+        # avoids a pandas FutureWarning about concatenating empty/all-NA entries.
```
delucchi-cmu:
```suggestion
        # avoids a pandas FutureWarning about concatenating empty frames.
```

delucchi-cmu (changes requested):
Agree with @hombit - though the warning indicates "empty/all-NA", we don't explicitly test for the "all-NA" case. Just removing that part of the sentence should be ok.

dougbrn (commented):
Agree that this looks like a reasonable fix, pending clarification on the all-NA margin. Just not claiming that per @delucchi-cmu is also totally fine.

delucchi-cmu (approved):

PR #1477, opened 2026-07-15
Fix mypy pre-commit failure

### LINKED ISSUE #1478: pre-commit's mypy hook is failing
**Bug report**

After operations merge, pandas-stubs was introduced, which makes `mypy` more strict about pandas-related typing. We should fix these issues so `pre-commit` CI is not red anymore, and local pre-commit is not failing all the time.

hombit:
There are a lot of mypy failures after `operations` were merged, this is my attempt of fixing them.
Fixes #1478


dougbrn (approved):
Thanks for doing this, looks good!

PR #1482, opened 2026-07-15
Handle nested field index search.

delucchi-cmu:
See also https://github.com/astronomy-commons/hats-import/pull/702

- Adds handling of the dataframe portion of the `search_points`, when the indexed column is a nested column.
- Catalog collection that has both a base column and nested column index


dougbrn (approved):
Jumping in to try to move this along for v0.10.1. This looks good to me and I'm going to go ahead and merge it!

Issue #1457, opened 2026-07-06
Meta issue: cases when `Catalog.write_catalog()` produces an invalid catalog


hombit:
**Bug report**

There are a lot of small issues with `write_catalog` when the resulting on-disk catalog cannot be opened with LDSB, or provides incorrect metadata. I use this issue to list the ones I am aware of, because it may be useful to aggregate the fixes into a single PR.

### Properties referring to columns

While it is ok to assume that healpix and coordinate columns didn't change (but see below), we should inspect other properties mentioning column names and remove those which were excluded from the catalog.

**`hats_cols_sort`**

AFAIK we don't check if user changed it, so it is probably easier to just remove this property.

Reproducible example: 
```python
lsdb.open_catalog("s3://ipac-irsa-ztf/ztf/enhanced/dr24/lc/hats/").write_catalog("<PATH>")
```
This would not load non-default '_healpix_19', '_healpix_9' columns, but the result hats.properties would still have 
`hats_cols_sort=['_healpix_29', '_healpix_19', '_healpix_9']`

**`hats_cols_survey_id`**

This column may be deleted by user, reproducible example:
```python
lsdb.open_catalog("s3://ipac-irsa-ztf/ztf/enhanced/dr24/lc/hats/", columns=[]).write_catalog("<PATH>")
```

### Dataset without coordinate columns

**`hats_col_ra`** and **`hats_col_dec`**

Currently, it is pretty easy to produce a HATS catalog which would fail on most of the LSDB's API calls by removing coordinate columns:

```python
import lsdb

# Still loads coordinates
cat = lsdb.open_catalog("s3://nasa-irsa-euclid-q1/contributed/q1/merged_objects/hats", columns=["object_id"])
cat[["object_id"]].write_catalog("/tmp/euclid-q1-oid")
cat_oid = lsdb.open_catalog("/tmp/euclid-q1-oid")
# Fails with KeyError: "['ra', 'dec'] not in index"
cat_oid.cone_search(0, 0, 1)
```

Please also note that `hats_col_ra` and `hats_col_dec` are still there. Similar problems would occur when renaming ra/dec columns.

One of the possible solutions is re-crating ra/dec on the fly from the healpix column when available. Another solution is fixing #1307.

**Environment Information**

<details>
<summary>INSTALLED VERSIONS</summary>

```
Paste the output from `lsdb.show_versions()`
```

</details>

<details>
<summary>Traceback</summary>

```
[...]

Issue #1361, opened 2026-05-05
Add more args to write_catalog


delucchi-cmu:
**Feature request**

There are many flags in `hats_import.ImportArguments` that allow for customization of the catalog, and runtime. These should be available in LSDB as well (and some might already be):

- `create_metadata`: Create /dataset/_metadata parquet from all data partitions.
- `create_per_partition_stats` : Create per_partition_statistics.parquet, based on footers from all data partitions.
- `drop_empty_siblings`: when determining bins for the final partitioning, should we keep result pixels at a higher order (smaller area) if the 3 sibling pixels are empty.
- `row_group_kwargs`: additional keyword arguments to use in creation of rowgroups when writing files to parquet.
- `should_write_skymap`: main catalogs should contain skymap fits files
- `write_table_kwargs`: additional keyword arguments to use when writing files to parquet (e.g. compression schemes).
- `columns`: a list of columns to write, that should be a subset of catalog's columns. (new - just nice to have)

csestili:
Arg | Status
------|------
row_group_kwargs | implemented and tested
create_parquet_metadata | already implemented, added test
create_per_partition_statistics | already implemented, added test
should_write_skymap | already implemented, added test
write_table_kwargs | already implemented, added test
drop_empty_siblings | blocked, need help with design
columns | not started

csestili:
thinking about the best way to implement `drop_empty_siblings` in `write_catalog()` (or really, in `to_hats()`).

`drop_empty_siblings` is currently implemented when loading a catalog from a DataFrame, in this hierarchy:  
`lsdb.loaders.dataframe.from_dataframe()`   
`lsdb.loaders.dataframe.from_dataframe.DataframeCatalogLoader`   
`lsdb.loaders.dataframe.from_dataframe.DataframeCatalogLoader.load_catalog()`  
`lsdb.loaders.dataframe.from_dataframe.DataframeCatalogLoader._compute_pixel_list()`  
`hats.pixel_math.partition_stats.generate_alignment()`

So it seems like we need a way to apply `generate_alignment()` across an already-loaded catalog.

csestili:
For `lsdb.Catalog.write_catalog(..., row_group_kwargs, ...)`, I'm not sure exactly how row groups are controlled while writing parquet files.

Option A: 
Some row group arguments, like `row_group_size`, can just be passed to `write_table_kwargs`, and they eventually get passed to `df.write_parquet()`.

Option B:
[...]

Issue #71, opened 2023-11-21
Keep column metadata/provenance for joined catalogs


smcguire-cmu:
Having extra metadata with the columns for a joined catalog to track which tables they came from.

delucchi-cmu:
Additional points from meeting:

- when performing additional operations, only require the added prefix/suffix if *required* for disambiguation
- if we need to extend/change the `_hipscat_index`, we can add metadata about the version it was generated with and determine compatibility matrix
- we *could* include support for VOField (e.g. VOTable-as-parquet) metadata in this effort

This could do with a design doc for defining the scope.

Issue #31, opened 2023-10-02
Add provenance column for cross-match/join results


delucchi-cmu:
e.g. for Tape Single Pixel, we need a column to indicate if the source is coming from `ztf` or `pan_starrs`

PR #1488, opened 2026-07-17
Augment write_catalog arguments

### LINKED ISSUE #1377: `write_catalog` wrongly inherits properties from the "parent" catalog
**Bug report**

When I was running analysis with ZTF DR23 (S3 catalog from IPAC), I've noticed, that `write_catalog` reuses some properties from the original catalog, see some suspicious rows in the output `hats.properties` I've got:

```properties
dataproduct_type=source
hats_npix_suffix=/
hats_creator=Caltech/IPAC-IRSA
hats_order=7
```

Please notice that the output dataset does not use "leaf folders" as `hats_npix_suffix` suggests. I'm not sure about `hats_order=7` because the only partition in the output I have is order 4, but maybe it is correct in some way...

**Environment Information**

Main branch hats and LSDB

<details>
<summary>INSTALLED VERSIONS</summary>

```
--------      SYSTEM INFO      --------
python        : 3.14.4
python-bits   : 64
OS            : Linux
OS-release    : 4.18.0-553.22.1.el8_10.x86_64
Version       : #1 SMP Wed Sep 11 18:02:00 EDT 2024
machine       : x86_64
processor     : x86_64
byteorder     : little
LC_ALL        : 
LANG          : 
--------   INSTALLED VERSIONS   --------
lsdb          : 0.9.1.dev81+ge21caf0ad
hats          : 0.9.1.dev10+g3017cc408
nested-pandas : 0.6.9
pandas        : 2.3.3
numpy         : 2.4.4
dask          : 2026.3.0
pyarrow       : 24.0.0
fsspec        : 2026.4.0
```

</details>

**Before submitting**
Please check the following:

- [x] I have included information about my environment, including the version of this package (e.g. `lsdb.show_versions()`)

delucchi-cmu:
This change effects several open issues. 

--------------

Closes #1377

- dataproduct_type: addressed with picking the best catalog type when you merge (and you still have both catalog types). this uses the ordering from the catalog type enumeration, and seems to make enough sense (e.g. object or source? source; map or object? object; source or association? source). note that we still have issues related to provenance tracking for crossmatch results in #31 and #71 (single digit issues - woah)
- hats_npix_suffix: was being passed by the wrong name
- hats_creator: added flag on `write_catalog:inherit_provenance` that tries to clear out known provenance-related fields. i can imagine a case where someone has been keeping their own provenance fields, and would be sad to have to set them every time they wrote out a derived data product.
- hats_order/moc_sky_fraction: updated with the new partition info

Remaining on #1457
- properties referring to columns / coordinate columns

--------------

Related to (but does not close) #1361

Added:
- create_metadata (but renamed to create_parquet_metadata, because loads of OTHER metadata gets written)
- create_per_partition_stats 
- should_write_skymap
- write_table_kwargs (to better distinguish all the kinds of kwargs)

NOT added:
- drop_empty_siblings: write_catalog just uses whatever partitioning the catalog already has, so this doesn't make sense.
- row_group_kwargs: this would be a lot of work that i don't feel like doing today
- columns: same - don't need to add more things today

--------------

Extra changes:
- tried to be a little more consistent with ordering/passing arguments through
- tidied up some of association creation
- also have a change to hats to tidy up the float serialization (https://github.com/astronomy-commons/hats/pull/710)
- tried to switch long positional parameter lists to mostly optional parameters with a `*`

### DIFF
--- a/src/lsdb/io/to_association.py
+++ b/src/lsdb/io/to_association.py
@@ -233,7 +233,7 @@ def to_association(
         "contains_leaf_files": True,
         "hats_order": partition_info.get_highest_order(),
         "total_rows": int(np.sum(counts)),
-        "moc_sky_fraction": f"{partition_info.calculate_fractional_coverage():.5f}",
+        "moc_sky_fraction": f"{partition_info.calculate_fractional_coverage():.5}",
     } | new_provenance_properties(inherit_provenance=inherit_provenance)
 
     max_separation = np.max(max_separations)
--- a/src/lsdb/io/to_association.py
+++ b/src/lsdb/io/to_association.py
@@ -12,6 +12,7 @@ from hats.catalog import CatalogType, PartitionInfo, TableProperties
 from hats.catalog.catalog_collection import CatalogCollection
 from hats.pixel_math import HealpixPixel
 from upath import UPath
+from lsdb.io.common import new_provenance_properties
 
 from lsdb.io.common import set_default_write_table_kwargs
 
@@ -67,6 +68,8 @@ def to_association(
     *,
     base_catalog_path: str | Path | UPath,
     catalog_name: str | None = None,
+    inherit_provenance: bool = False,
+    addl_hats_properties: dict | None = None,
     primary_catalog_dir: str | Path | UPath | None = None,
     primary_column_association: str | None = None,
     primary_id_column: str | None = None,
@@ -75,9 +78,11 @@ def to_association(
     join_to_primary_id_column: str | None = None,
     join_id_column: str | None = None,
     separation_column: str | None = None,
+    create_parquet_metadata: bool = True,
+    create_per_partition_statistics: bool = True,
+    error_if_empty: bool = True,
     overwrite: bool = False,
-    addl_hats_properties: dict | None = None,
-    **kwargs,
+    write_table_kwargs: dict | None = None,
 ):
     """Writes a crossmatching product to disk, in HATS association table format.
     The output catalog comprises partition parquet files and respective metadata.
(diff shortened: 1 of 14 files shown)


src/lsdb/io/to_hats.py
```diff
@@ -397,7 +425,9 @@ def to_hats(
         total_rows=int(np.sum(counts)),
         default_columns=default_columns,
         hats_max_rows=hats_max_rows,
-        hats_npix_suffix=npix_suffix,
+        moc_sky_fraction=partition_info.calculate_fractional_coverage(),
+        hats_order=partition_info.get_highest_order() if len(partition_info) else 0,
```
hombit: Same question, should we skip it instead? Set to -1?

src/lsdb/loaders/dataframe/margin_catalog_generator.py
```diff
@@ -271,7 +271,7 @@ def _create_catalog_info(self, catalog_name: str | None = None, **kwargs) -> Tab
         if kwargs is None:
             kwargs = {}
         kwargs.pop("catalog_type", None)
-        kwargs = kwargs | new_provenance_properties(hats_estsize=0)
+        kwargs = kwargs | new_provenance_properties() | {"hats_estsize": 0}
```
hombit: Would it be better to just skip the field?

hombit (commented):
Do I understand correctly that `write_table_kwargs` change is technically a breaking change? If yes, we should highlight it in the release notes.

src/lsdb/io/to_association.py
```diff
@@ -211,8 +233,8 @@ def to_association(
         "contains_leaf_files": True,
         "hats_order": partition_info.get_highest_order(),
         "total_rows": int(np.sum(counts)),
-        "moc_sky_fraction": f"{partition_info.calculate_fractional_coverage():0.5f}",
-    }
+        "moc_sky_fraction": f"{partition_info.calculate_fractional_coverage():.5f}",
```
hombit: Maybe scientific notations for small values?
```suggestion
        "moc_sky_fraction": f"{partition_info.calculate_fractional_coverage():.5}",
```

src/lsdb/io/to_hats.py
```diff
@@ -397,7 +425,9 @@ def to_hats(
         total_rows=int(np.sum(counts)),
         default_columns=default_columns,
         hats_max_rows=hats_max_rows,
-        hats_npix_suffix=npix_suffix,
+        moc_sky_fraction=f"{partition_info.calculate_fractional_coverage():.5f}",
```
hombit: Same suggestion as above
```suggestion
        moc_sky_fraction=f"{partition_info.calculate_fractional_coverage():.5}",
```

hombit (approved):

PR #1508, opened 2026-07-22
Fix progress bars in write catalog

smcguire-cmu:
Ensures that during write_catalog calls we see one progress bar with the correct description


dougbrn (approved):
Nice, looks good!

PR #1509, opened 2026-07-22
Add DASK info clobber and simplest plotting standard

nevencaplar:
Adds a way to clobber DASK verbosity with info msgs, adds dashboard, adds few hints about plots


dougbrn (approved):
Nice, looks good

PR #1517, opened 2026-07-23
Fix failing build, use hardoced lightcurve; Add point about S3

nevencaplar:


dougbrn (commented):
Ah, the template notebook was the culprit! Apologies for not catching a potential memory issue in the review, I am tempted to suggest we just make this not run at all since it's purpose is just to be a template.

smcguire-cmu (approved):

PR #1519, opened 2026-07-24
Add more args to write_catalog

csestili:
See issue #1361 .

PR https://github.com/astronomy-commons/hats/pull/730 needs to be merged before the CI checks will pass, because I moved a function into hats.


csestili:
All the CI checks are failing because `import hats_import` is failing. I thought it was a dependency of lsdb, need to look into this.

docs/developer/tutorial_template.ipynb
```diff
@@ -120,6 +120,35 @@
     "plt.show()"
    ]
   },
+  {
```
csestili, line 123: I didn't make this change but I guess it got pulled in when I was merging

csestili:
@csestili : to finish this up, we need to:
- move the row group splitting to `hats` and call it from both `lsdb` and `hats-import` (DONE)
- check the parquet metadata num_row_groups---why isn't it being updated? (DONE)
- move drop_siblings and columns to their own PR

Will address these when I get back from leave, 8/14

tests/lsdb/io/test_to_hats.py
```diff
@@ -588,3 +588,140 @@ def test_save_catalog_no_summary_by_default(small_sky_order1_catalog, tmp_path):
...
+    # Norder=1/Dir=0/Npix=45: [10, 10, 9]
+    # Norder=1/Dir=0/Npix=46: [10, 10, 10, 10, 2]
+    # Norder=1/Dir=0/Npix=47: [10, 8]
+    for pix_num, expected_num_row_groups in [(44, 5), (45, 3), (46, 5), (47, 2)]:
+        pf = pq.ParquetFile(
+            base_catalog_path / f"small_sky_order1/dataset/Norder=1/Dir=0/Npix={pix_num}.parquet"
+        )
+        assert pf.num_row_groups == expected_num_row_groups
+
+    # NOTE why doesn't the number of row groups in the metadata change?
```
csestili, line 668: unsure why the number of row groups in the metadata is not changing. reviewers, any ideas?

docs/developer/tutorial_template.ipynb
hombit: Something is wrong here, we don't want unexpected changes. Could you please try to merge/rebase on origin/main?

src/lsdb/io/to_hats.py
```diff
@@ -634,3 +660,46 @@ def create_modified_catalog_structure(
     new_hc_structure.catalog_info = new_hc_structure.catalog_info.copy_and_update(**kwargs)
     new_hc_structure.catalog_info.catalog_name = catalog_name
     return new_hc_structure
+
+
+# Copied from hats_import.catalog.map_reduce
+# TODO DRY refactor
+def _split_to_row_groups(table, row_group_kwargs, pixel_order):
```
hombit: Should it go to `hats` first, and then re-used in both `hats-import` and `lsdb`?

src/lsdb/io/to_hats.py
```diff
@@ -634,3 +660,46 @@ def create_modified_catalog_structure(
...
+    -------
+    split_tables : list[pa.Table]
+        The input table split into row group chunks.
+    """
+    if "num_rows" in row_group_kwargs:
+        chunk_size = row_group_kwargs["num_rows"]
+        return [table.slice(i, chunk_size) for i in range(0, len(table), chunk_size)]
+    if "subtile_order_delta" in row_group_kwargs:
+        split_tables = []
+        parent_pixels = table[SPATIAL_INDEX_COLUMN].to_numpy()
```
hombit: What if the user has deleted the `_healpix_29` column, or another column is used as the healpix column? We should properly cover this case because 1) it is easy for the user to strip this column, 2) the HATS standard allows no healpix column, and 3) the HATS standard allows a custom name and healpix order for the healpix column.
But again, I think this should go to `hats` package first.

hombit (changes requested):

docs/developer/tutorial_template.ipynb
csestili: that makes sense. I probably messed something up when I was rebasing/merging in the latest changes. I'll try again.

src/lsdb/io/to_hats.py
csestili: yes, thank you for reminding me! I will do that.

src/lsdb/io/to_hats.py
csestili: yes, this makes sense! I was worried about the `_healpix_29` seemingly magic value. I'll add more robust handling for this.

csestili:
@csestili TODO:
- move `_split_to_row_groups()` into `hats`. (DONE)
- call `_split_to_row_groups()` from `lsdb` and `hats_import`. test the calls to make sure they still work. (DONE)
- strengthen handling of `_healpix_29` column assumption from `_split_to_row_groups()`. (DONE)
- figure out why metadata num_row_groups is not being changed and fix (DONE)
- keep all the other tests I wrote in lsdb

csestili:
rebased this branch onto main and force-pushed here, hopefully this cleans up the commits I didn't actually write

src/lsdb/io/to_hats.py
csestili: Moved to `hats`. See https://github.com/astronomy-commons/hats/pull/730

src/lsdb/io/to_hats.py
csestili: Moved to `hats` and added error handling. See https://github.com/astronomy-commons/hats/pull/730

docs/developer/tutorial_template.ipynb
csestili: rebased and it seems to have fixed this.

csestili:
I figured out why the parquet metadata num_row_groups seemed to be not changing. I was testing the wrong metadata file. Just switched it to the correct metadata file and updated tests accordingly, everything is working as it should be.

tests/lsdb/io/test_to_hats.py
csestili: resolved in https://github.com/astronomy-commons/hats/pull/730. I was testing the wrong metadata file. have updated to fix this.

csestili:
Just merged https://github.com/astronomy-commons/hats/pull/730 .

In this branch, pre-commit checks pass locally. CI checks on GitHub should pass once above PR gets integrated.

src/lsdb/io/to_hats.py
```diff
@@ -77,7 +82,20 @@ def perform_write(
...
+        with pq.ParquetWriter(
+            pixel_path.path,
+            table.schema,
+            filesystem=pixel_path.fs,
+            **kwargs,
+        ) as writer:
+            for table in rowgroup_tables:
+                writer.write_table(table)
+    else:
+        df.to_parquet(pixel_path.path, filesystem=pixel_path.fs, **kwargs)
```
hombit: Do we need to pass schema here, the same way we do above?

src/lsdb/io/to_hats.py
```diff
@@ -77,7 +82,20 @@ def perform_write(
         npix_suffix=npix_suffix,
         npix_parquet_name=npix_parquet_name,
     )
-    df.to_parquet(pixel_path.path, filesystem=pixel_path.fs, **kwargs)
+    if row_group_kwargs:
+        table = pa.Table.from_pandas(df)
```
hombit: I guess we should do `df.to_pyarrow()` instead? I'm afraid that something wrong may go here with list vs large list arrays. It would probably produce "right" result, but we may accidentally cast types back and forth.

tests/lsdb/io/test_to_hats.py
hombit: Could you please add a test, which checks that `row_group_kwargs=something` and `row_group_kwargs=None` produces the same result schema?

hombit (approved):

src/lsdb/io/to_hats.py
csestili: Hmm, I don't think there is a df.to_pyarrow() method, is there? `pa.Table.from_pandas(df)` is the only way I could find to convert pandas df to pyarrow table, lmk if I'm missing anything!

hombit: `NestedFrame` has `.to_pyarrow`, which is aware about this large_list/list thing.
https://github.com/lincc-frameworks/nested-pandas/blob/a5599a4df62185efb4d19b0bff42bf0a913eafd6/src/nested_pandas/nestedframe/core.py#L2358

I think it is missed from the API docs - sorry

PR #1548, opened 2026-08-04
Refresh the API surface figure and automate its upkeep

nevencaplar:
Clickable API map on landing page updated to match new version of LSDB. Claude also suggested a number of improvment to automate the creation in the future, via regen_api_surface_png.py and check_api_surface.py - and I can not even pretend that I understand what is happening.


dougbrn (approved):
Seems good to me, there's a lot of noise from claude in the actual explanations which I don't think is necessary, but given that this is just a dedicated tool offshoot in our repo I think it's okay.

Issue #1522, opened 2026-07-24
Core + Extension Catalogs: Partial Join Requirements


dougbrn:
Sub-Issue of #1521 

When a core catalogue is joined against an extension covering <100% of objects (e.g., cutouts above a magnitude/size threshold, spectra excluding bad-flagged entries), the goal is that there are zero false matches and zero dropped core rows; unmatched core objects should be retained with null extension columns.

Issue #1521, opened 2026-07-24
Core + Extension Catalogs: Tracker Issue


dougbrn:
This ticket tracks the larger effort to properly support Core + Extension catalogs in LSDB/HATS. This should have a design document with a clearer set of requirements.

Sub-Issues:
* Hats issue: https://github.com/astronomy-commons/hats/issues/721
* Partial Join Target Behavior: #1522
* Extension Loading API: #1523 

When loading a Core catalog, there should be dedicated API functions that load an extension layer. Potentially with some discovery method for extensions, but that could be a later effort.

PR #1552, opened 2026-08-11
adds regression coverage for sparse nested left joins across independently backed storage sources

EmployeeNo427:
This is test-only coverage for existing nested-join behavior. The sparse catalog is an Extension stand-in.

1. The tests use:

* A Core catalog stored on the local filesystem.
* A physically sparse HATS catalog stand-in stored in `memory://` and reopened through LSDB.
* Missing pixels and missing rows within covered pixels.
* Duplicate scientific rows represented in main and margin partitions.
* An object with the correct identity placed in a different overlapping pixel.
* A declared data partition removed after catalog creation to verify that the resulting read failure propagates.

2. What is verified:

* Every Core row is preserved exactly once, including its original order and index.
* Exact 0:N nested matches are attached to the expected Core objects.
* Unmatched Core rows receive null nested values.
* Duplicate main/margin representations do not produce duplicate matches.
* Sparse scientific coverage is treated as valid noncoverage.
* Identity matching remains partition-local under the current `join_nested` behavior.
* A missing declared partition raises `FileNotFoundError` instead of being silently interpreted as scientific noncoverage.
* The in-memory fixture explicitly sets `scheduler=None` so LSDB uses its in-process fallback in this no-client test environment. This does not exercise multiprocessing.

3. Testing:

* Focused join test suite: 20 passed
* `DASK_SCHEDULER=processes` environment-isolation check: 4 passed; the fixture resets the scheduler to `None`, so this does not exercise the process scheduler
* Full Linux, Windows, lowest-dependency, pre-commit, and benchmark workflows passed

4. Questions that need clarifications:

* Is this prerequisite regression coverage useful to merge before the Extension design is finalized?
* Should the displaced-identity case remain as an explicit characterization of the current partition-local behavior?


nevencaplar:
Hi ZJ (@EmployeeNo427)! Thank you for your contribution. Before we spend more valuable human time to dig more into this, can you please add more human-readable instructions and information. In particular:
1. What is the motivation for this work, and does it connect to any tickets.
2. What is the addition provided in this work, connecting it to the motivation.
3. Explain in more detail what is meant by ``The sparse catalog is an Extension stand-in``.

EmployeeNo427:
> Hi ZJ (@EmployeeNo427)! Thank you for your contribution. Before we spend more valuable human time to dig more into this, can you please add more human-readable instructions and information. In particular:
> 
> 1. What is the motivation for this work, and does it connect to any tickets.
> 2. What is the addition provided in this work, connecting it to the motivation.
> 3. Explain in more detail what is meant by `The sparse catalog is an Extension stand-in`.

Thank you for looking into this! @nevencaplar I hope I am not messing things up.

If I understand the project correctly, an Extension may contain data for only part of the Core catalog. [LSDB issue 1522](https://github.com/astronomy-commons/lsdb/issues/1522) gives examples such as image cutouts selected by magnitude or size, and spectra from which bad-flagged entries have been excluded. It says the goal is “zero false matches and zero dropped core rows” and that “unmatched core objects should be retained with null extension columns.” That is the behavior I am trying to test in this PR. [HATS issue 721](https://github.com/astronomy-commons/hats/issues/721) also says that Extension catalogs should be readable from a different source rather than requiring them to be stored together with their Core catalogs. Therefore, the tests use a Core catalog and a separately stored, physically sparse HATS catalog, then verify that Core rows are preserved, available rows are matched correctly, unmatched Core rows receive null values, and storage failures remain visible.

By “Extension stand-in,” I mean a temporary substitute used for testing. I could not find a finalized `ExtensionCatalog` type or contract in the current LSDB and HATS default branches. [HATS issue 721](https://github.com/astronomy-commons/hats/issues/721) gives a conceptual description of an Extension catalog, while [LSDB issue 1521](https://github.com/astronomy-commons/lsdb/issues/1521) proposes dedicated APIs for loading Extension layers. For that reason, I constructed an ordinary HATS catalog that behaves like a possible future Extension only in the ways relevant to this test: it is stored separately, contains data for only some Core objects, and is joined to the Core catalog at query time.

This does not say that the stand-in represents the final Extension format. It tests whether today’s `join_nested` behavior can safely handle the kind of sparse, separately stored data that a future Extension may contain. Once the design is finalized, I can update.

Admittedly, this PR only supports [LSDB 1521](https://github.com/astronomy-commons/lsdb/issues/1521), [1522](https://github.com/astronomy-commons/lsdb/issues/1522) and [HATS issue 721](https://github.com/astronomy-commons/hats/issues/721), but it does not directly implement or close any of them.

tests/lsdb/catalog/test_join.py
```diff
@@ -378,49 +384,166 @@ def test_join_nested(small_sky_catalog, small_sky_order1_source_with_margin):
         )
 
 
-def test_join_nested_how_left(small_sky_order1_catalog, small_sky_order1_source_with_margin, helpers):
```
dougbrn, line 381: Is there a reason why you've opted to overwrite an existing test? I understand this test is closely related, but my assumption would have been that this PR would introduce it's own set of tests while preserving the existing tests.

dougbrn (commented):
Hi @EmployeeNo427 , appreciate you putting this PR together. I think what's here seems like a good first step, I just have a structural question, below.

tests/lsdb/catalog/test_join.py
EmployeeNo427: Hi @dougbrn, thank you for the review. I initially replaced the existing test because I viewed the new case as a more specific version of the same left-join scenario. Looking at them side by side, however, they exercise different inputs: the existing test uses a `pixel_search` subset, while the new test uses a separately stored, physically sparse catalog, so I agree that the new coverage should be added alongside the existing test rather than replace it. I’ll restore the original test and keep the new scenario as a separate test.

nevencaplar:
hi @EmployeeNo427 - is this ready for re-review after your changes? It would be helpful if you could request re-review by clicking on the circle button in the top-right corner, once when you are ready: 
image

dougbrn (approved):
This looks good to me! There are two mypy failures in the pre-commit CI, but they look unrelated to this PR. I'll go ahead and merge this, as I'm not sure whether you have the ability to hit the merge button yourself

PR #1559, opened 2026-08-13
Print missing ranges in the catalog statistics

camposandro:
In the Catalog representation, when the statistics are requested, a missing min/max value for a pixel was represented with the placeholder, `...`. This PR prints missing values as they come from `per_partition_statistics` (e.g. a float column with missing stats will now show `nan`). The placeholder is still shown for pixels and columns that have no statistics at all.


nevencaplar (approved):
Looks good to me.

PR #1564, opened 2026-08-18
`to_lance` fix for S3

maxwest-uw:
While working through [this](https://github.com/astronomy-commons/hats-cloudtests/issues/145) issue in `hats-cloudtests` I discovered that the `to_lance` function doesn't actually work with UPaths that point to cloud storage. This PR ensures that a UPath that points to an S3 location will actually maintain and map the appropriate information so that `lancedb` can write to it.

Note that this currently only handles the mapping for S3 buckets, it should eventually work for GCS as well but I didn't have a way to test those changes locally :s

Will also have an additional PR in `hats-cloudtests`


src/lsdb/io/to_lance.py
```diff
@@ -12,6 +12,71 @@
...
+
+    if protocol == "s3":
+        return _map_s3_storage_options(fsso)
+
+    # Unknown/unsupported protocol (http, memory, ...) -- nothing sensible
+    # to translate; let lance/lancedb use its own defaults.
+    return None
+
+
+def _map_s3_storage_options(fsso: dict) -> dict[str, str]:
```
csestili, line 48: looks good, just add a docstring that calls out the details you show in your tests. e.g. top-level takes precedence over client, http is only allowed when endpoint url is http, etc.

csestili (changes requested):
Just add a docstring, but otherwise looks good to me! You might want to get a second opinion from someone who has more experience with the lsdb S3 integration.

csestili (approved):
Looks good to me!

PR #1573, opened 2026-08-27
Update DP1 photoz notebook

camposandro:
Re-ran the DP1 photoz notebook and made some small tweaks. E.g., some cells still showed the outputs for the first version of the photoz data (before March 2026), and there were no instructions on how to install `qp`.


nevencaplar (approved):
Looks good to me.

PR #1574, opened 2026-08-28
Combine both catalog names in join()/join_nested() default output name

### LINKED ISSUE #397: Output catalog name from joined catalog keeps only left catalog name
should be something more descriptive, similar to how crossmatch is a_x_b

agu2347:
## Summary

Fixes #397.

`Catalog.join()` and `Catalog.join_nested()` defaulted `output_catalog_name` to just the left catalog's name (`self.hc_structure.catalog_info.catalog_name`) when `output_catalog_name` was not explicitly provided. This silently drops the right catalog's name from the result, unlike `crossmatch()`/`crossmatch_nested()` and `merge_asof()`, which already combine both catalog names into the default name (crossmatch uses `"{left_name}_x_{right_name}"`, as documented in its own docstring).

## Fix

- Changed the default `output_catalog_name` in `join()` and `join_nested()` (`src/lsdb/catalog/catalog.py`) to `f"{self.name}_x_{other.name}"`, matching the existing crossmatch convention.
- Updated the docstrings for both methods to document the new default (`str, default {left_name}_x_{right_name}`), matching how crossmatch's docstring already documents its default.

## Tests

Added `test_small_sky_join_default_output_catalog_name` and `test_join_nested_default_output_catalog_name` in `tests/lsdb/catalog/test_join.py`, which check both the new default name and that an explicitly passed `output_catalog_name` is still respected.

## Verification

- Ran the two new tests — pass with the fix.
- Reverted the fix locally (restoring the old left-only-name default) and reran the same two tests: both fail, reproducing the exact bug reported in the issue (e.g. `assert 'small_sky' == 'small_sky_x_small_sky_order1'`).
- Reran the surrounding join/join_nested/association-join tests in `tests/lsdb/catalog/test_join.py` (`test_join_wrong_columns`, `test_join_wrong_suffixes`, `test_small_sky_join_small_sky_order1`, `test_small_sky_join_overlapping_suffix`, `test_join_nested`, `test_join_nested_how_left`, `test_join_nested_invalid_how`, `test_join_association`, `test_join_association_overlapping_suffix`) to confirm no other behavior changed.

Signed-off-by: agu2347 <agu2347@users.noreply.github.com>

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -1073,7 +1073,7 @@ class Catalog(HealpixDataset):
             between pixels and individual rows.
         suffixes : tuple[str,str]
             Suffixes to apply to the columns of each table
-        output_catalog_name : str
+        output_catalog_name : str, default {left_name}_x_{right_name}
             The name of the resulting catalog to be stored in metadata
         suffix_method : str, default "all_columns"
             Method to use to add suffixes to columns. Options are:
@@ -1151,7 +1151,7 @@ class Catalog(HealpixDataset):
             )
 
         if output_catalog_name is None:
-            output_catalog_name = self.hc_structure.catalog_info.catalog_name
+            output_catalog_name = f"{self.name}_x_{other.name}"
 
         new_catalog_info = create_merged_catalog_info(
             self,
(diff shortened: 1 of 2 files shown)


nevencaplar:
Hi! Thank you for your contribution. Unfortunately, we don't accept unsolicited agent-generated PRs, regardless of patch quality. Reviewing agent-authored PRs from people with no prior contact with the project costs considerable maintainer time If you'd like to contribute, comment on an issue first and agree the approach with us.

PR #1575, opened 2026-08-29
Fix join() output catalog name to combine both catalog names

### LINKED ISSUE #397: Output catalog name from joined catalog keeps only left catalog name

arbelonson-source:
Fixes #397.

`Catalog.join()` and `Catalog.join_nested()` defaulted `output_catalog_name` to the left catalog's name only, so a join result carried no trace of the right catalog. This changes the default to `f"{left}_join_{right}"`, following the convention already used by `crossmatch` (`_x_`) and `merge_asof` (`_merge_asof_`).

Covers the plain join path, the association (`through=`) path, and `join_nested()`. An explicit `output_catalog_name=` still takes precedence.

## Testing

`tests/lsdb/catalog/test_join.py`: 25 passed, including 4 new tests. I confirmed the new tests actually catch the bug by reverting the source change — 3 of the 4 fail with the old left-only name.

## Notes

- I used `_join_` spelled out rather than reusing crossmatch's `_x_` infix, so join results stay visually distinct from crossmatch results. Easy to switch if you'd rather be consistent with `_x_`.
- This changes the default output name, which is the behavior change #397 asks for.
- `merge_map()` has the same left-only-name pattern, but I left it alone as out of scope here. Glad to follow up separately if you want it changed too.

## Disclosure

This change was prepared with AI assistance (Claude Code). The tests above were run locally.

### DIFF
--- a/src/lsdb/catalog/catalog.py
+++ b/src/lsdb/catalog/catalog.py
@@ -1073,7 +1073,7 @@ class Catalog(HealpixDataset):
             between pixels and individual rows.
         suffixes : tuple[str,str]
             Suffixes to apply to the columns of each table
-        output_catalog_name : str
+        output_catalog_name : str, default {left_name}_join_{right_name}
             The name of the resulting catalog to be stored in metadata
         suffix_method : str, default "all_columns"
             Method to use to add suffixes to columns. Options are:
@@ -1151,7 +1151,7 @@ class Catalog(HealpixDataset):
             )
 
         if output_catalog_name is None:
-            output_catalog_name = self.hc_structure.catalog_info.catalog_name
+            output_catalog_name = f"{self.name}_join_{other.name}"
 
         new_catalog_info = create_merged_catalog_info(
             self,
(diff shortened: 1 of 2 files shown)


nevencaplar:
Hi! Thank you for your contribution. Unfortunately, we don't accept unsolicited agent-generated PRs, regardless of patch quality. Reviewing agent-authored PRs from people with no prior contact with the project costs considerable maintainer time If you'd like to contribute, comment on an issue first and agree the approach with us.

arbelonson-source:
Understood, and thanks for saying so plainly rather than leaving it open.

That's a fair policy and I should have checked how you prefer to be approached before opening a PR — the disclosure was there, but disclosing after the fact still spends your review time, which was the thing you're protecting. I won't reopen or re-submit this.

For anyone finding this later: #397 is still open, and the change was a two-line default in `Catalog.join()` to name the output `{left}_join_{right}`, matching the existing `_x_` and `_merge_asof_` conventions.

Issue #1391, opened 2026-05-18
More flexible guess at radec columns for from_dataframe


delucchi-cmu:
Add possibility for looking for "ra_*" and "dec_*", but only within the first three columns of the catalog.

could add logic to : lsdb/src/lsdb/loaders/dataframe/dataframe_catalog_loader.py:_find_column

nevencaplar:
The goal is that we catch names for ra and dec columns in realistic catalogs, if the user did not provide them, such as

Gaia-like      │ source_id, ra_deg, ra_error, dec_deg, dec_error
LSST/Butler    │ objectId, coord_ra, coord_dec, …         
VizieR         │ RAJ2000, DEJ2000, Vmag    (note that this is different than DEC!)
PS1 MeanObject │ objID, raMean, decMean

Note that original ticket above is not general enough to catch all of these examples. I would also expand to search for positional columns within first 4 columns - We dont want to search beyond to not accidentally include various other possible columns with ra or dec in their names which are not actual main positional coordinates.

nevencaplar:
Also, avoid regression: Exact ra/dec must keep resolving at any position — for instance ["id","band","mjd","flux","flux_err","ra","dec"] should still find ra/dec colums

PR #1581, opened 2026-09-01
Improve user experience for ra/dec parsing and error reporting

csestili:
This PR addresses:
- https://github.com/astronomy-commons/lsdb/issues/1393
- https://github.com/astronomy-commons/lsdb/issues/1391

Changes:
- Error messages about ra/dec column issues are more descriptive (done)
- Search process that finds ra/dec columns will be expanded to include more patterns from known catalogs (done)
- When ra/dec assumptions are violated during map_partitions(), errors are reported during the map_partitions() call, not hidden until later (done for single_partition. for general case, moving this to another issue.)


src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -410,6 +424,7 @@ def map_partitions(
         if isinstance(new_cat._operation, MapPartitions) and not new_cat._operation.is_df_type:
             col_name = new_cat.columns[0]
             return new_cat.to_dask_dataframe()[col_name]
+        # NOTE unsure how to do the ra/dec checks above in this case
```
csestili: @smcguire-cmu could you take a look at this and advise? I think it would be a good idea to check that map_partitions() doesn't change the ra/dec columns, so that this assumption violation gets caught when the calculation happens, reducing downstream errors. See the simple little implementation above for `compute_single_partition==True`.

I'm unsure how the dask operations work, so I'm not sure how to implement this for the full case when `compute_single_partition==False`. What do you think?

smcguire-cmu: Yeah, that's a nice feature! I think for the dask version of it the check would need to happen inside the function submitted to dask, so that it's run on the client during the computation. I think the best way to do this would be to write a function that wraps the users function and checks the output in the same way you do above. We do a similar thing when [verifying the meta in operations here](https://github.com/astronomy-commons/lsdb/blob/f2d1fef5a607a729d35fea245aa003b31fce4148/src/lsdb/operations/lsdb_ops.py#L21). You can just raise and error the same way and it'll bubble up through dask. If you can use the same wrapping tools like the above example too, it'll make it so the function the user sees on the dashboard is the same that they wrote instead of the wrapper function.

csestili: I see, I'll give that a shot, thank you!

csestili:
Updated search process:
- case-insensitive matches for 'ra' or 'dec' anywhere in column list are matched
- in the first 4 columns, column names are checked against a dict of known ra/dec column names. I added ra/dec columns for all sources currently present in [lsdb.data.io](https://data.lsdb.io/)
- in the first 4 columns, a heuristic search finds column names that contain (case-insensitive) 'ra' or 'dec', as long as they don't contain 'err', 'sig', 'std', 'var'

Error messages have been updated to inform the user how to supply ra/dec column names.

src/lsdb/catalog/dataset/healpix_dataset.py
csestili: all right, I looked into this, and I think I'm going to kick this work to its own issue. I'm interested to spend some time with someone more experienced 👀 to see a good implementation

tests/lsdb/catalog/test_catalog.py
```diff
@@ -1062,6 +1065,36 @@ def divme(df, _=None):
...
+    # Should raise because ra/dec column names change
+    with pytest.raises(ValueError, match=re.escape("'source_ra' not found in result. map_partitions() must not change names of ra or dec columns 'source_ra', 'source_dec'.")):
+        small_sky_source_catalog.map_partitions(rename_cols, ['source_ra', 'source_dec'], ['my_ra', 'my_dec'], compute_single_partition=True)
+
+    # Should raise because map_partitions() changes ra/dec values
+    # for col_name in ['source_ra', 'source_dec']:
+    #     with pytest.raises(ValueError, match=re.escape(f"ra/dec values have changed. map_partitions() must not change values of ra or dec columns 'source_ra', 'source_dec'.")):
+    #         small_sky_source_catalog.map_partitions(my_evil_function, col_name, compute_single_partition=True)
+
+    with pytest.raises(ValueError, match=re.escape(f"ra/dec values have changed. map_partitions() must not change values of ra or dec columns 'source_ra', 'source_dec'.")):
```
csestili: This test fails because no error is raised, I can't figure out why....

csestili: fixed it!

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -143,14 +171,39 @@ def _find_column(self, search_term: str) -> str:
...
+                if str(col_name).lower() in RADEC_COLUMN_MAPPING[search_term]:
+                    matches.append(col_name)
+                # heuristic match
+                elif _is_radec_like(str(col_name).lower(), search_term):
+                    logging.warning(
+                        "Warning: heuristic match found for `%s`: '%s'. Please check correctness!",
+                        search_term,
+                        col_name,
+                    )
+                    matches.append(col_name)
```
smcguire-cmu, line 193: If I'm reading this right, I think if there is an exact match in a column and also a heuristic match in the first 4 columns it will error. I think separating this to first check an exact match, then do the column search on the first 4 columns would make more sense to me.

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -347,3 +400,26 @@ def _generate_moc(self):
...
+        "err",
+        # Variance / standard deviation terms
+        "sig",
+        "std",
+        "var",
+    ]
+
+    if any(t in col_name for t in negative_terms):
+        return False
+    return any(t in col_name for t in RADEC_COLUMN_MAPPING[search_term])
```
smcguire-cmu, line 425: Checking any substring match seems a little overly accepting to me. E.g. "ra" in "parallax" would return true. If it's already been discussed with a scientist and they think it's fine you can ignore me, but maybe starting with the string, or bounded on either side by a space/underscore/hyphen might be a better check.

src/lsdb/catalog/dataset/healpix_dataset.py
```diff
@@ -385,17 +385,35 @@ def map_partitions(
...
             partition_cat = self.partitions[partition_index]
             pixel = partition_cat.get_healpix_pixels()[0]
             partition = partition_cat.compute()
+            orig_coords = partition[[ra_col, dec_col]]
             result = (
                 func(partition, pixel, *args, **kwargs) if include_pixel else func(partition, *args, **kwargs)
             )
             if not isinstance(result, pd.DataFrame):
                 return result
+            # Check that ra and dec columns are still present
```
smcguire-cmu, line 402: It might also be worth checking that if the index is still a healpix index that it matches the radec cols

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
csestili: That's right, actually I included a test case for this: [link](https://github.com/astronomy-commons/lsdb/pull/1581/changes#diff-53af8c02d9efd5d3b90158ed37b8ec7fa1ece14c48a8420b663e3e0a7adca6e3R608). I was thinking, if there's an exact match as well as a heuristic match, the user may be unintentionally passing multiple columns with ra/dec values, and we want to error to prevent any issues.

I see what you mean though, and maybe this is too defensive. Maybe if there's an exact match, we should just finish the search there. What do you think? (cc @nevencaplar)

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
csestili: That makes sense, I'll make this check more precise as you recommended. What do you think (@nevencaplar )? Note that the known matches include e.g. "RAJ2000", "DEJ2000", and we can always extend the list of known matches if the heuristic match doesn't catch a column name in the future.

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
csestili: Here's a similar case when there's ambiguity following a known match: [link](https://github.com/astronomy-commons/lsdb/pull/1581/changes#diff-53af8c02d9efd5d3b90158ed37b8ec7fa1ece14c48a8420b663e3e0a7adca6e3R558).

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
csestili: all right, I made the check more precise and updated tests to match. I think you were right, I feel better about it this way. thank you!

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
smcguire-cmu: Ah yeah, that's a good point! I'd be interested to hear what the scientists think is most useful in cases like this

csestili: Ok, after discussion, I'm going to keep this how it is for now. Thanks for the review!

src/lsdb/catalog/dataset/healpix_dataset.py
csestili: Thanks, just added this check, what do you think?
https://github.com/astronomy-commons/lsdb/pull/1581/changes/063248e0a4eb690e2cfa5b0e6310e05a217f7005#diff-be091568272b8d0d3a31c4a06391864892595a87de4578e075e160defb899d28R418

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -143,14 +171,39 @@ def _find_column(self, search_term: str) -> str:
...
+                matches.append(col_name)
+            # attempt column search only in the first 4 columns
+            elif col_idx < 4:
+                # known matches
+                if str(col_name).lower() in RADEC_COLUMN_MAPPING[search_term]:
+                    matches.append(col_name)
+                # heuristic match
+                elif _is_radec_like(str(col_name).lower(), search_term):
+                    logging.warning(
+                        "Warning: heuristic match found for `%s`: '%s'. Please check correctness!",
```
smcguire-cmu: I think I remember a suggestion to reword this error to explain what heuristic match means a bit more

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
```diff
@@ -143,14 +171,39 @@ def _find_column(self, search_term: str) -> str:
...
-            raise ValueError(f"No column found for {search_term}")
+            raise ValueError(
+                f"No column found for '{search_term}' (required). You can supply ra/dec column names "
+                "using the arguments `ra_column`, `dec_column`."
+            )
         if n_matches > 1:
-            raise ValueError(f"Found {n_matches} possible columns for {search_term}")
+            raise ValueError(
+                f"Found {n_matches} possible columns for '{search_term}': {matches}. Please "
+                "rename columns to disambiguate."
```
smcguire-cmu: Maybe suggest setting the columns via the arguments rather than renaming?

smcguire-cmu (approved):
Looks good! Just a few questions about error messages that I think were brought up at the last demo, but the healpix index check seems good! I'm not sure about what's happening with the test coverage, it looks like it's saying a bunch of lines aren't covered that seem like they should be in the tests, maybe it's a bug in the reporting?

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
csestili: Thank you, fixed!

src/lsdb/loaders/dataframe/dataframe_catalog_loader.py
csestili: Thank you, fixed!

csestili:
@smcguire-cmu thanks for the re-review! I changed the error messages to be more informative.

I'm going to wait until I can get someone else's eyes in person to check the codecov issues, before merging.

delucchi-cmu (approved):

PR #1584, opened 2026-09-03
Filtering cross-match stream

hombit:
Proof-of-concept for the multi-cross-matching stream, doing 'left' joins with possible skipping of some left catalogs.

### DIFF
--- a/src/lsdb/streams/catalog_streams.py
+++ b/src/lsdb/streams/catalog_streams.py
@@ -260,14 +260,13 @@ class CatalogIterator(Iterator[pd.DataFrame]):
         return int(np.ceil(len(self.partitions_left) / self.iterable.partitions_per_chunk))
 
 
-class CrossMatchStream(CatalogStream):
+class CrossMatchStream(InfiniteStream):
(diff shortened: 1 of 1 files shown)


Smith42:
I hit an error with skipped partitions here, copying "diagnosis" below:

When CrossMatchStream decides to skip a pixel, it builds a bare self.catalog.partitions[[pixel]].map_partitions(...) graph instead of a crossmatch graph — for ~6% of skip-branch pixels, that graph's pixel_to_key_map doesn't contain the pixel it was asked to build (likely because DESI's own partition tree subdivides that particular sky area more finely than the crossmatch alignment's pixel set, so the coarser pixel key doesn't exist in the raw partition-selection graph).

Smith42:
Looks like it is working! But we still get an odd batch composition (see blue new vs green old)

Screenshot 2026-09-03 at 15 41 28
Screenshot 2026-09-03 at 15 41 59

I don't know if this is an issue yet, but it is worth flagging I think as this suggests a different xmatching ordering.

PR #1590, opened 2026-09-03
Add fetched-pair outer crossmatch (how='outer')

Smith42:
lsdb's crossmatch supported 'inner' and 'left' at the row level. This adds how='outer' to Catalog.crossmatch and the top-level crossmatch function, recovering unmatched primary-right rows from the pixel pairs the left alignment already fetches and discards.

This is an outer join, including the union of left and right catalogs.

Internals:                                                                                                                                                              
   - Outer crossmatching now uses `PixelAlignmentType.OUTER`, so every pixel from either catalog is visited and the result is partitioned on the union of both catalogs.   
   - `CrossmatchArgs` gains `right_native_mask`, marking which rows of the primary-right partition belong to the aligned pixel, so margin-cache rows are not emitted as    
 independent outer rows and coarse right partitions are not duplicated across finer aligned pixels.                                                                        
   - `CrossmatchArgs` also gains `left_native_len`, distinguishing native left rows from neighborhood rows. Neighborhood matches are emitted by their home-pixel task and  
 elsewhere only prevent their right rows from being emitted unmatched.                                                                                                     
   - `AlignAndApply` pixel slots may contain tuples of pixels, allowing right-only tasks to gather and concatenate multiple adjacent left partitions.                      
   - Boundary planning uses the ring one HEALPix order finer than each right-only aligned pixel, built with `PixelTree.from_healpix`; aligned pixels existing only in the  
 right margin halo are dropped.                                                                                                                                            
   - `_create_crossmatch_df` is refactored into matched/left-only/right-only blocks so inner, left, and outer share one assembly path; algorithms with no extra columns    
 are supported.                                                                                                                                                            
   - `how` is now validated explicitly (`'inner' | 'left' | 'outer'`); unknown values previously fell through to inner-like assembly.                                      
                                                                                                                                                                           
   The initial fetched-pair implementation was verified end-to-end against the real DESI x LegacySurvey stream (6,703 rows: 1,403 matched + 4,333 left-only + 967          
 recovered image-only, zero loss/duplication), motivating the design. The current branch supersedes that limited behavior with exact full outer semantics.                 
                                                                                                                                                                           
   Tests: `test_crossmatch_outer_join.py` covers matched and unmatched rows on both sides, both suffix methods, right-only sky coverage, boundary and cross-pixel matches, 
 multi-partition neighborhood gathering, left-margin exclusion of matched rows, coarse-right/finer-aligned-pixel deduplication, native-mask validation, margin and radius  
 requirements, no-extra-column custom algorithms, and invalid `how`. Full suite: 469 passed, 23 skipped. `black`/`isort`/`pylint`/`mypy` clean.


src/lsdb/catalog/catalog.py
```diff
@@ -262,13 +267,16 @@ def crossmatch(
         """Perform a cross-match between two catalogs
 
         The pixels from each catalog are aligned via a `PixelAlignment`, and cross-matching is
-        performed on each pair of overlapping pixels. The resulting catalog will have partitions
-        matching an inner pixel alignment - using pixels that have overlap in both input catalogs
-        and taking the smallest of any overlapping pixels.
+        performed on each selected pixel pair. Inner joins use overlapping pixels; left joins
+        use the left catalog's coverage; outer joins use the union of both.
```
dougbrn, line 269: My understanding from the scope description is that outer uses lefts pixel alignment only, but this seems to imply that we are also including the right catalogs coverage, am I misunderstanding or is this incorrect?

dougbrn: There's a few more mentions claiming this below as well

src/lsdb/operations/functions/crossmatch_catalog_data.py
```diff
@@ -308,6 +386,118 @@ def perform_crossmatch_nested(
...
+        )
+    tree_order = alignment.pixel_tree.tree_order
+    intervals = np.empty((0, 2), dtype=np.int64)
+    if len(pixel_mapping):
+        orders = pixel_mapping[PixelAlignment.ALIGNED_ORDER_COLUMN_NAME].to_numpy(dtype=np.int64)
+        pixels = pixel_mapping[PixelAlignment.ALIGNED_PIXEL_COLUMN_NAME].to_numpy(dtype=np.int64)
+        shift = 2 * (tree_order - orders)
+        intervals = np.stack([np.left_shift(pixels, shift), np.left_shift(pixels + 1, shift)], axis=1)
+    alignment = PixelAlignment(
+        PixelTree(intervals, tree_order), pixel_mapping, alignment.alignment_type, alignment.moc
```
dougbrn: It's probably not safe to pass the moc along directly here, as we've pruned the margin only pixels, so we need an updated moc to reflect that

src/lsdb/operations/functions/crossmatch_catalog_data.py
```diff
@@ -308,6 +386,118 @@ def perform_crossmatch_nested(
...
+
+
+def _plan_outer_alignment(
+    left: Catalog, right: Catalog, alignment: PixelAlignment
+) -> tuple[PixelAlignment, list[tuple[HealpixPixel, ...] | None]]:
+    """Drop margin-only pixels from an OUTER alignment and plan boundary partitions."""
+    pixel_mapping = alignment.pixel_mapping
+    right_only = pixel_mapping[PixelAlignment.PRIMARY_ORDER_COLUMN_NAME].isna()
+    if right_only.any():
+        primary_right_pixels = {(p.order, p.pixel) for p in right.get_healpix_pixels()}
```
dougbrn: nit: native_right_pixels may be a more appropriate name to reflect the non-margin nature of these pixels, primary is a left-catalog loaded term.

src/lsdb/operations/functions/crossmatch_catalog_data.py
```diff
@@ -308,6 +386,118 @@ def perform_crossmatch_nested(
...
+    if not right_only_mask.any():
+        return boundary_lists
+    rings: dict[tuple[int, int], list[tuple[int, int]]] = {}
+    ring_cells: set[tuple[int, int]] = set()
+    for i in np.flatnonzero(right_only_mask):
+        row = pixel_mapping.iloc[i]
+        order, pixel = int(row[PixelAlignment.ALIGNED_ORDER_COLUMN_NAME]), int(
+            row[PixelAlignment.ALIGNED_PIXEL_COLUMN_NAME]
+        )
+        ring = [(order + 1, cell) for cell in get_margin(order, pixel, 1)]
```
dougbrn, line 439: get_margin here uses a fixed `delta_order=1`, which does not take into account a defined `radius_arcsec`. In `_validate_outer_crossmatch` margin thresholds are cognizant of `radius_arcsec`. We might need to account for cases where delta_order=1 is insufficient to cover the full radius, and instead compute delta_order here as a function of radius_arcsec and pixel size at the order. For transparency, this was flagged by claude, and I'm not entirely sure how real the threat of missing matches is in practice.

dougbrn (commented):
This looks like it's moving in a good direction, I have a few comments, not all of which I feel require a direct fix for this to be mergable

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
```diff
@@ -96,21 +96,49 @@ def crossmatch(
             l_inds = np.array([], dtype=np.int64)
             r_inds = np.array([], dtype=np.int64)
             extra_cols = self.extra_columns.copy() if self.extra_columns is not None else pd.DataFrame()
+        elif how == "outer" and (crossmatch_args.left_df is None or len(crossmatch_args.left_df) == 0):
+            # Outer-only: a right-only pixel with no nearby left rows matches nothing. Every
+            # primary right row is emitted unmatched, and there are no left rows to emit.
+            l_inds = np.array([], dtype=np.int64)
+            r_inds = np.array([], dtype=np.int64)
+            extra_cols = self.extra_columns.copy() if self.extra_columns is not None else pd.DataFrame()
```
hombit, line 99: The code looks identical to the previous if branch

src/lsdb/catalog/catalog.py
Smith42: This is an update after a conversation with Kostya about left alignmement vs outer join! I will update the PR text in a moment to clarify the scope

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
```diff
@@ -96,21 +96,49 @@ def crossmatch(
...
                 "Crossmatch algorithm must return left and right indices and extra columns with same length"
             )
+        left_df = crossmatch_args.left_df
+        right_matched_mask = None
+        if how == "outer" and crossmatch_args.left_native_len is not None:
+            native_len = crossmatch_args.left_native_len
+            visitor = l_inds >= native_len
+            if visitor.any():
+                right_matched_mask = np.zeros(len(crossmatch_args.right_df), dtype=bool)
+                right_matched_mask[r_inds] = True
```
hombit: Could this be replaced with `right_matched_mask = np.where(r_inds)[0]`?

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
```diff
@@ -96,21 +96,49 @@ def crossmatch(
...
+                l_inds = l_inds[~visitor]
+                r_inds = r_inds[~visitor]
+                extra_cols = extra_cols[~visitor].reset_index(drop=True)
```
hombit: Why do we drop the index only in this case?

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
```diff
@@ -241,116 +269,95 @@ def _create_crossmatch_df(
...
-                        right_join_part,
-                    ],
-                    axis=1,
-                ),
+        n_unmatched = len(out) - len(extra_cols)
+        assert n_unmatched >= 0, (
+            f"Logic error: extra_cols ({len(extra_cols)} rows) cannot exceed output ({len(out)} rows). "
+            "This indicates a bug in the crossmatch algorithm or the join logic."
+        )
+        if n_unmatched:
```
hombit: Please do it explicitly
```suggestion
        if n_unmatched > 0:
```

src/lsdb/core/crossmatch/crossmatch.py
```diff
@@ -77,9 +81,10 @@ def crossmatch(
         The name of the output catalog.
     require_right_margin : bool, default False
         Whether to require a right margin.
-    how: str
-        How to handle the crossmatch of the two catalogs.
-        One of {'left', 'inner'}.  Defaults to 'inner'.
+    how : {'inner', 'left', 'outer'}, default 'inner'
+        How to assemble the crossmatch. ``outer`` also emits unmatched right rows,
+        scanning sky covered only by the right catalog; it requires margin caches on
+        both catalogs (left threshold at least the matching radius, right at least twice).
```
hombit: Let's describe all of them, not just outer

hombit (commented):
1. I think we don't have a solid interface in the existing LSDB code for some of the stuff we are doing here. It may be extremely confusing that when we have no matches on the left catalog, the left catalog's ra and dec columns are reused. @Smith42, @dougbrn, do you think that we should provide new columns for `ra` and `dec` when doing outer/right joining?
2. I have also noticed that the result is not necessarily sorted by `_healpix_29`, which we normally would expect if the inputs were sorted. It is kinda not an issue for inner/left join, but here we would probably want to sort by it explicitly.
3. From the code I cannot really understand where we handle the left margins, but from both PR tests and my own testing it looks like it is being taken into account somehow. While I believe that this PR scope is good, after it is merged we should work on follow-ups to support 1) `join(how="outer")` and 2) `how="right"` for both join and crossmatch.

src/lsdb/catalog/catalog.py
Smith42: updated :)

src/lsdb/operations/functions/crossmatch_catalog_data.py
Smith42: Commit [a95e362](https://github.com/astronomy-commons/lsdb/pull/1590/commits/a95e362c1bae483957ccb5f2a67ca0e26097b6c6) resolves this

src/lsdb/operations/functions/crossmatch_catalog_data.py
Smith42: Resolved with [b79ee05](https://github.com/astronomy-commons/lsdb/pull/1590/commits/b79ee05805c540b278ed5f84cf199b72279e1576)

src/lsdb/operations/functions/crossmatch_catalog_data.py
Smith42: I'm not sure how we could fix this. Would it be a problem in the "typical" use case of a ~1 arcsec xmatch boundary? I think this will only become an issue for partition order > 16

hombit:
I have just realized that I've missed a lot of code in `operations` during the review.

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
Smith42: Removed in [6bb8a70](https://github.com/astronomy-commons/lsdb/pull/1590/commits/6bb8a70400b6b7ef9959cd64fa0b542d4f8d721e)

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
Smith42: We didn't need to do that, fixed with [5a9df27](https://github.com/astronomy-commons/lsdb/pull/1590/commits/5a9df275382d552609f6622438688efa5a2342b0)

src/lsdb/operations/functions/crossmatch_catalog_data.py
dougbrn: Yeah I think you're correct in saying that this only becomes a possible issue around order 16+, for the typical use case. There is rough math we could do to try to dynamically set this, but I don't neccesarily know the best path forward on that myself, perhaps @hombit does. It might be suitable to just have a comment here summarizing the limitation/potential for missed matches?

src/lsdb/core/crossmatch/crossmatch.py
Smith42: Done with [e3ea2cc](https://github.com/astronomy-commons/lsdb/pull/1590/commits/e3ea2cc8e2c98a6ab15b49c2d8c5c7206607c14a)

Smith42:
> 1. I think we don't have a solid interface in the existing LSDB code for some of the stuff we are doing here. It may be extremely confusing that when we have no matches on the left catalog, the left catalog's ra and dec columns are reused. @Smith42, @dougbrn, do you think that we should provide new columns for `ra` and `dec` when doing outer/right joining?

Yes I think this could cause some confusion -- would it be enough to document that we follow what is done for the inner join for the matched pairs, and then use the right coordinates only for the unmatched right rows? Or do we need this somehow within the HATS spec?

> 2. I have also noticed that the result is not necessarily sorted by `_healpix_29`, which we normally would expect if the inputs were sorted. It is kinda not an issue for inner/left join, but here we would probably want to sort by it explicitly.

Hmm, would this significantly add to compute cost for the xmatch? There's a tradeoff we need to decide on there. Happy to sort the healpix_29 column if you feel it is worth the added compute.

> 3. From the code I cannot really understand where we handle the left margins, but from both PR tests and my own testing it looks like it is being taken into account somehow. While I believe that this PR scope is good, after it is merged we should work on follow-ups to support 1) `join(how="outer")` and 2) `how="right"` for both join and crossmatch.

Yep we can scope this out as a follow up

hombit:
> 1.

I think we should go with new ra/dec columns. I propose to discuss on the Friday 1pm ET HATS meeting.

> 2. I have also noticed that the result is not necessarily sorted by `_healpix_29`, which we normally would expect if the inputs were sorted. It is kinda not an issue for inner/left join, but here we would probably want to sort by it explicitly.

What is usually expensive is data copy with shuffling, not sorting itself. We are already shuffling the partitions when building these blocks, and theoretically sorting should not add much compared to cross-matching, especially if we are doing it on the level of our temporary indexes. I think we should change the "block" approach you use to build blocks from the positional indexes first, sort them by `_healpix_29`, and only then bring all other columns with one-time data copy per left/right frames.

BTW, this PR doesn't really handle the possible situation in which left and right catalogs have different healpix columns of different orders (which is allowed by the HATS standard). I think I would just (re)construct `_healpix_29` from the result ra/dec columns in the case when we don't have `_healpix_29` for both tables.

hombit:
@Smith42 we discussed it and we think that for outer/right joins we would add new `_ra`, `_dec` columns to handle new coordinates, with the same implementation you have: left catalog's coordinates for "inner" and "left" matches, and right catalog coordinates for the rest. What do you think about it?

Smith42:
added new columns for _ra and _dec in outer join with [296a815](https://github.com/astronomy-commons/lsdb/pull/1590/commits/296a81533d1b79c8440c2a66b2a275886eeacad4)

Smith42:
@hombit I had a go at the healpix reordering in [c7679d9](https://github.com/astronomy-commons/lsdb/pull/1590/commits/c7679d99909d3d9c2de40880275ffdede3e0355a) but happy to revert and follow your lead on this as I am not sure exactly what spec you had in mind.

src/lsdb/operations/functions/crossmatch_catalog_data.py
hombit: @smcguire-cmu please help!

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
```diff
@@ -336,30 +332,37 @@ def _create_crossmatch_df(
...
+        if how == "outer" and left_coordinates is not None and right_coordinates is not None:
+            for output_coord, left_coord, right_coord in zip(
+                ("_ra", "_dec"), left_coordinates, right_coordinates, strict=True
+            ):
+                if left_coord is not None and right_coord is not None:
+                    left_values = cast(pd.Series, out[left_name_map[left_coord]])
+                    right_values = cast(pd.Series, out[right_name_map[right_coord]])
+                    out[output_coord] = left_values.combine_first(right_values)
+            ra_values = cast(list[float], cast(pd.Series, out["_ra"]).to_numpy())
+            dec_values = cast(list[float], cast(pd.Series, out["_dec"]).to_numpy())
```
hombit, line 362: I'm lost with these `cast`s, especially the last two. `cast` doesn't actually change the type, it just provides the type checker a hint that we actually know which of the possible types our variable is. So here we convert the series to a numpy array and immediately tell the type checker that it is actually a list, which cannot be true!  
I believe it would be better to just turn type checks off for these lines instead.

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
```diff
@@ -336,30 +332,37 @@ def _create_crossmatch_df(
...
+            spatial_index = compute_spatial_index(ra_values, dec_values)
```
hombit, line 363: Another option is changing type hints on this function, but it goes out of scope of this PR

hombit (commented):
Thanks, @Smith42! I'm asking @smcguire-cmu to look over the margin cache code and to approve it

src/lsdb/operations/functions/crossmatch_catalog_data.py
smcguire-cmu: So I looked over the code and I think I'd like to suggest a different approach to doing the margins this way. As far as I can tell, this code runs when we have a right pixel with no matching left pixels, and we need to find the left rows that might match with them. It does technically work to generate the margin ring and concat all the matching left partitions, but that can really blow up the dask graph with concatting an unbounded number of left partitions together and could result in large memory requirements for that concat task and makes efficiently scheduling those tasks a lot harder. (It is an edge case to be fair for there to be no left partition but lots of neighboring left pixels, but e.g. matching near edges of coverage it could happen). Especially when we do already have that in the left margin catalog. The margin catalogs include the 'negative pixels' which cover this exact case, so including the left margin catalog in the alignment I think means the whole boundary pixel lists and concat tasks in AlignAndApply become unnecessary. We would need to extend the pruning to exclude left margin only partitions too, but I think it's a cleaner approach and doesn't lead to the whole task graph issues.

It doesn't actually solve the order 16+ margin problem, just kicks it to a different place. In margin catalog generation we basically do the same technique to figure out the margin pixels and we only support a ring of one healpix pixel width, so if there was a catalog at order 17 and we wanted a 1 arcsec margin we just wouldn't be able to generate it, but yeah as you mention that's a pretty big edge case and we already don't support it.

src/lsdb/core/crossmatch/abstract_crossmatch_algorithm.py
smcguire-cmu: Yeah we should change that function to ArrayLike, can we make an issue for that in hats?

PR #1616, opened 2026-09-15
Pandas 3 Compatibility

dougbrn:
In process of debugging this, found that the installation step in our CI was prematurely failing to resolve the dependencies if a valid match existed on hats main, because we were attempting an install from the dev branch first. Workflows updated to fix this


delucchi-cmu (approved):

hombit (approved):
Thanks

